PR #5, opened 2015-06-15
relax peerDependencies.

chirag04:
cannot install this with react-native 0.5.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,6 @@
   },
   "license": "MIT",
   "peerDependencies": {
-    "react-native": "^0.4.0"
+    "react-native": ">= 0.4.4"
   }
 }


oblador:
Thanks!

chirag04:
no problem. can you bump the version?

oblador:
Was just about to, done!

chirag04:
sweet. Thanks :)

Also, can you have a quick look at my last comment on this?: https://github.com/oblador/react-native-keychain/issues/3

chirag04:
What do you think about just dropping peerDependency and having it as a dependency? 

It works with 0.5 but wont work with 0.6 now.

https://github.com/chirag04/react-native-dashed-border/blob/master/package.json#L24-L26

PR #7, opened 2015-08-12
adds 0.9.0-rc to package.json

bleonard:


bleonard:
not exactly the best situation, but making it work for my current code

PR #38, opened 2016-10-27
typescript defenitions added. Callback interface fixed to match de fa…

dryganets:
- Typescript definitions added.
- Fixed callbacks signature.


oblador:
Thanks!

Issue #35, opened 2016-09-27
[iOS] SecItemAdd() -- The dictionary does not specify a value for kSecAttrAccessible key


aprct:
It's setting off some flags for sanity checks in our company's static code security scanner:

> **Attack Vector:** SecItemAdd
> **Description:** In this call to SecItemAdd(), the dictionary specified does not appear to include a value for the key kSecAttrAccessible. This value defaults to kSecAttrAccessibleWhenUnlocked, which may result in keychain data being accessible and unlocked longer than necessary.
> **Remediation:** Review the potential values for kSecAttrAccessible, and ensure that the one used in this application meets the security needs of its users.
> **References:** [CWE](http://cwe.mitre.org/data/definitions/522.html)  [Apple Keychain Ref](https://developer.apple.com/reference/security/1658642-keychain_services/1663541-keychain_item_accessibility_cons)

I know it's an issue of semantics, but it would be nice if this setting was either exposed in the JS module, or set explicitly to the default value: [https://developer.apple.com/reference/security/ksecattraccessiblewhenunlocked](https://developer.apple.com/reference/security/ksecattraccessiblewhenunlocked)

Thanks :)

oblador:
Accessibility option is now supported in [1.1.0](https://github.com/oblador/react-native-keychain/releases/tag/v1.1.0).

PR #39, opened 2016-10-28
keychain records are always accessible on iOS.

dryganets:
if it's not - keychain can't be accessed on locked devices if push notification comes in

in our app we can't access encrypted database in result in we are getting push in background.


christianlundkvist:
This PR solves the problem that @dryganets is having, but the downside is that all users would be using the lower security setting that their keychain items are available on the lock screen. I would suggest making the PR a bit more general by adding the `kSecAttrAccessible` attribute as a user input to the high-level functions instead. 

Then the user can select their own level of `kSecAttrAccessible`. I was myself looking at using a stronger setting for this value, so this change would kill both of our birds with one stone :)

christianlundkvist:
This would also solve Issue https://github.com/oblador/react-native-keychain/issues/35

PR #51, opened 2017-01-10
Fix accessible

alexanderjarvis:
Every now I then I get reports of `Keychain item not found` for users that should have a keychain item.

I suspect that the default is not good enough and have always previously set `AccessibleAfterFirstUnlock`.

This changes the default so that the keychain item is available when the app is in the background, or even as I suspect, in a transition between background or locked state.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,18 +10,6 @@ type SecAccessible =
   | 'AccessibleAfterFirstUnlockThisDeviceOnly'
   | 'AccessibleAlwaysThisDeviceOnly'
 
-function convertError(err) {
-  if (!err) {
-    return null;
-  }
-  if (Platform.OS === 'android') {
-    return new Error(err);
-  }
-  var out = new Error(err.message);
-  out.key = err.key;
-  return out;
-}
-
 /**
  * Saves the `username` and `password` combination for `server`.
  * @param {string} server URL to server.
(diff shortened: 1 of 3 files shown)


oblador:
This is a potentially sensitive change, could you make this configurable instead of changing it for everybody?

alexanderjarvis:
Sure but if I make it configurable, I'd still recommend it as the default.

oblador:
Okay!

index.js
```diff
@@ -1,19 +1,42 @@
...
+type SecAccessible =
+  | 'AccessibleWhenUnlocked'
+  | 'AccessibleAfterFirstUnlock'
+  | 'AccessibleAlways'
+  | 'AccessibleWhenPasscodeSetThisDeviceOnly'
+  | 'AccessibleWhenUnlockedThisDeviceOnly'
+  | 'AccessibleAfterFirstUnlockThisDeviceOnly'
+  | 'AccessibleAlwaysThisDeviceOnly'
+
+function convertError(err) {
```
oblador: This function is unused AFAIK, please remove :-)

oblador:
Neat, thanks a bunch! If you could add the type script definitions too that would be just spendid, but no biggie if not.

alexanderjarvis:
@oblador hey can you merge and release this soon?

PR #54, opened 2017-02-07
iOS Keychain Sharing 

maxperry:
Hi,

I split the index.js file into index.ios.js and index.android.js, and I refactored all the methods of the iOS module (except the ones for Shared Web Credentials) to allow to pass an accessGroup string. 

This enables Keychain sharing on iOS, which is very convenient when developing an App Extension, just like in my case where I needed to access the keychain in order to share the login credentials. 

Please refer to the links below for more info:
http://evgenii.com/blog/sharing-keychain-in-ios/
http://swiftandpainless.com/ios8-share-extension-with-a-shared-keychain/

We'll also need to update the README if you decide to merge.

Cheers!

### DIFF
--- a/RNKeychainManager/RNKeychainManager.m
+++ b/RNKeychainManager/RNKeychainManager.m
@@ -75,20 +75,28 @@ void rejectWithError(RCTPromiseRejectBlock reject, NSError *error)
   return reject(codeForError(error), messageForError(error), nil);
 }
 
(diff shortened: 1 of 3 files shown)


oblador:
Thanks! I wanted a slightly different API for this, but it's now released in [1.1.0](https://github.com/oblador/react-native-keychain/releases/tag/v1.1.0). Use it with an option object instead `{ accessGroup: 'com.my.app' }`.

PR #60, opened 2017-03-17
Update TypeScript definitions

markdaws:
This pull request updates the package.json to use the "types" attribute to declare where the type definition file is located. I am not sure if the "typescript" attribute was an older method of declaring this but I am using TypeScript 2.2.1 and the current type definitions are not found until I use the "types" attribute. See: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html for clarification.

I also updated some of the type definitions to have more accurate promise values.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   "typescript": {
     "definition": "typings/react-native-keychain.d.ts"
   },
+  "types": "typings/react-native-keychain.d.ts",
   "repository": {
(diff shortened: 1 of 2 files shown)


oblador:
💯 Thanks a lot!

oblador:
Released in [1.2.0](https://github.com/oblador/react-native-keychain/releases/tag/v1.2.0).

PR #65, opened 2017-05-03
protect the data stored in keychain by TouchId or Passcode

steffsan:
We added a couple of functions to protect the data stored into the keychain via TouchId or Passcode. One can check the local authentication settings on the device - if protected storage is possible. One can read and write securely saved items to the keychain. 
We added additional functions to keep the code separate and not cause any unexpected behaviour for existing code.
iOS only changes


oblador:
Wow, that's a nice PR. I want to give this a thorough review, but I'm really busy right now so it might take a while. Apologizing in advance 😃 

Any reason the deployment target is changed? Will this break any backwards compatibility?

Would it be possible to implement similar functionality for android too?

What's the advantage to this approach as opposed to using this library together with [`react-native-touch-id`](https://github.com/naoufal/react-native-touch-id)?

steffsan:
Hi, well, I am an iOS developer, I implemented those features on demand for some RN-developers, so I am afraid I cannot answer if something similar is possible on Android as well. 

"deployment target is changed" - well, I guess that was during experimentation, I got an error that I do not have the correct entitlements - on the web this was a well known error which is known to occur with earlier iOS versions - therefore I switch the deployment target. However, I think that this error should no longer occur. I think it was due to 'kSecAttrAccessGroup' and the either-or-relation to 'kSecAttrAccessControl' for an item - this caused the confusion.

"What's the advantage to this approach as opposed to using this library together with react-native-touch-id?" Well, it seems to be convenience mainly, I guess. Some of the RN-devs around me use the combination and they say it is a bit cumbersome. Nonetheless, it seems to work. I guess for that reason the other RN-devs requested if we can implement it directly, as the keychain supports it. Therefore we implemented it, let them test it for, awaited feedback .....

steffsan:
Hi, we checked again - I am afraid the code I wrote is primarily compatible to iOS 9 and later. Before there where different attributes, they changed that. So, I you want to keep the library compatible with older versions if iOS I have to look into that.

lefloh:
> What's the advantage to this approach as opposed to using this library together with react-native-touch-id?

You can't do anything wrong ;-) The secret is protected by the OS and must be unlocked by the user. 
Also this approach has an automatic fallback to Code if Touch ID fails or is not supported.

michaelgmcd:
Hey @steffsan @oblador, any status update on this? Looking into react-native solutions for Touch ID protected keychain access and this seems to be the only solution out there. Would prefer not to have to write a bridge myself.

Rovack:
@oblador This sounds like it would be an extremely useful feature for React Native apps to start supporting, and one that react-native-keychain would be best suited to offer.

It looks like it's been over 4 months since this PR was opened, so any chance it could be reviewed/merged?

Robophil:
Any update on this guys? This looks good.

markdboyd:
@oblador - I'll try to answer your questions, though I am somewhat new to this space myself so others should feel free to correct me if I'm wrong.

> What's the advantage to this approach as opposed to using this library together with react-native-touch-id?

My understanding is that `react-native-touch-id` implements TouchID as a means of local authentication only, meaning that it is checking that the fingerprint provided matches one enrolled on the device. Apparently, this local authentication check is [somewhat vulnerable to spoofing by an attacker who could modify the check at runtime](https://books.nowsecure.com/secure-mobile-development/en/ios/implement-touch-id-properly.html#details). 

However, the code in this PR is using TouchID/biometrics as a secure flag on the access control for a Keychain item containing user credentials, thus requiring a TouchID authentication anytime that data is requested from the keychain. This approach enables some more advanced login flows, such as "Remember me" type logins with TouchID. Furthermore, the internal methods used by the Keychain for verifying TouchID are not as exposed or vulnerable to spoofing as with local authentication. 

>Would it be possible to implement similar functionality for android too?

The Android Keystore documentation suggests that it is possible to [require user authentication to access stored keys](https://developer.android.com/training/articles/keystore.html#UserAuthentication).

[And this is a line from a code sample](https://github.com/googlesamples/android-AsymmetricFingerprintDialog/blob/master/Application/src/main/java/com/example/android/asymmetricfingerprintdialog/MainActivity.java#L190) showing the method that is used to add fingerprint authentication protection to your stored keys.

oblador:
@markdboyd Cool, thanks for your answers. I'll have a look at android and see if I can get a similar feature there.

oblador:
This is released in [3.0.0-rc.1](https://github.com/oblador/react-native-keychain/releases/tag/3.0.0-rc.1], android is currently on hold.

PR #81, opened 2017-08-13
Fix path for iOS module linking via CocoaPods

8ofproject:
Make installing manual more friendly for beginners


oblador:
No matter what I write someone is complaining because they all have the Podfile in different places 😅 Let's change for now :-)

PR #90, opened 2017-11-29
Upgraded Gradle build tools version to 26.0.2

Elethier:
When I was importing this library into my project, I noticed that the Gradle version and the build tools version on Android were too low to work with my project. I updated the version number to be compatible with Android Studio 2.3.3, and I have confirmed that this works with Android Studio 3.0. Android Studio 2.3.3 and 3.0 both require a minimum tools version of 25.0.0. 

I haven't seen any bugs in the library due to this version bump, so I thought I'd submit a pull request. Please let me know if there's any reason to keep the build tools version number at the current version.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -4,7 +4,7 @@ buildscript {
   }
 
   dependencies {
-    classpath 'com.android.tools.build:gradle:1.1.3'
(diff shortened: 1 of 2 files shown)


oblador:
Superseded by https://github.com/oblador/react-native-keychain/pull/139

PR #118, opened 2018-03-27
Clear form to better demonstrate "load"

TechyTimo:
Once the form is cleared, it make easier to observe that the credentials are getting loaded from the Keychain system...


oblador:
Makes sense, thanks!

PR #122, opened 2018-03-27
Update Typings for Option.accessible and others

jpmonette:
Adding the following enums & keys:
- SecAccessControl
- SecAccessible
- LAPolicy
- `accessible` for Options

PR #131, opened 2018-04-20
Add options param to getInternetCredentials typing

Un3qual:
I think `getInternetCredentials` should have an options param, but the Typescript typings file doesn't have it.


Un3qual:
Whoops didn't see existing PR #122 for this. Can y'all merge that one soon?

PR #133, opened 2018-06-19
Catch uncaught exceptions to prevent crash

maxkomarychev:
`NullPointerException` thrown on some Huawei devices is crashing entire app.
Handle all exceptions and rethrow them as caught exceptions.

This is an attempt to prevent issue described in https://github.com/oblador/react-native-keychain/issues/115.


maxkomarychev:
close in favor of #134

PR #134, opened 2018-06-19
Catch uncaught exceptions to prevent crash

maxkomarychev:
`NullPointerException` thrown on some Huawei devices is crashing entire app.
Handle all exceptions and rethrow them as caught exceptions.

This is an attempt to prevent issue described in https://github.com/oblador/react-native-keychain/issues/115.


oblador:
In lack of objections from @pcoltau I'm merging this.

maxkomarychev:
@oblador thanks!
can we expect new release including this any few other fixes to be published soon? Last RC was issued in the end of february

Issue #162, opened 2018-11-11
Why don't you use flow.js?


retyui:
When I looked through your repository, I noticed that you have both the Flow.js on the project and the command for the test it!
https://github.com/oblador/react-native-keychain/blob/8f33c0f75032aa7309e4cb574299fc4785076927/package.json#L7

But there is not a single file with comments `// @flow` 

 It's a joke?

PR #163, opened 2018-11-11
Add flow anotation

retyui:


retyui:
https://github.com/oblador/react-native-keychain/issues/162

.flowconfig
```diff
@@ -29,7 +29,23 @@ node_modules/react-native/flow-github/
 [options]
 emoji=true
 
+esproposal.optional_chaining=enable
+esproposal.nullish_coalescing=enable
+
 module.system=haste
```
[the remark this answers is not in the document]
retyui: No, we shouldn't. Because of all `react-native` imports based on haste system. 
You will understand when changed this and run test!

index.js
```diff
@@ -10,50 +11,36 @@ export const ACCESSIBLE = {
...
 
-export const BIOMETRY_TYPE = {
+export const BIOMETRY_TYPE = Object.freeze({
   TOUCH_ID: 'TouchID',
   FACE_ID: 'FaceID',
   FINGERPRINT: 'Fingerprint',
-};
+});
+
+export type SecAccessible = $Values<typeof ACCESSIBLE>;
```
vonovak: `$Values<typeof ACCESSIBLE>` is the same as saying `string`. The accurate version would be `export type SecAccessible = 'AccessibleWhenUnlocked' | 'AccessibleAfterFirstUnlock' ... | 'AccessibleAlwaysThisDeviceOnly'`. Not sure what the best way to go about this is... Any suggestions?

vonovak:
@oblador I think this does improve things, and imho can be merged right away

oblador:
@vonovak What about your comment? We fix it separately?

index.js
retyui: You are partly right, but `$Values`  works perfectly with frozen objects:

image

image

vonovak: interesting, I didn't know that. I guess that solves this.

vonovak:
@oblador that part is not a problem, as discussed https://github.com/oblador/react-native-keychain/pull/163#discussion_r242822781

PR #167, opened 2018-11-13
iOS deleteAll() functionality

markrickert:
This PR adds functionality for deleting all keychain data associated with your app. This functionality is iOS only for now (anyone want to tackle this on Android?)

Also adds a button to the demo app showing usage and makes the demo app reference the node module from the cloned repo. instead of `^3.0.0`


oblador:
I'm a bit hesitant to merge this as other third party libraries may be using the keychain for unrelated things like analytics or authentication. Can you think of a way of preventing that?

markrickert:
I totally understand the hesitation. Maybe just a big red warning in the docs or something?  I mostly use this in `__DEV__` mode to reset my app state.

...the person who makes the hammer isn't responsible if someone hits their thumb with said hammer 🙃

KeychainExample/App.js
```diff
@@ -69,6 +69,19 @@ export default class KeychainExample extends Component {
     }
   }
 
+  async deleteAll() {
```
koyta, line 72: `deleteAll = () => {...}` and call it without creating an arrow function in render body `onPress={deleteAll}`

oblador:
This should now be possible to achieve using from [`getAllGenericPasswordServices()`](https://github.com/oblador/react-native-keychain#getallgenericpasswordservices) from https://github.com/oblador/react-native-keychain/pull/396

PR #199, opened 2019-02-20
Make a separate build folder for Linux, instead of a separate build/lib folder

PombeirP:
I'm getting a weird situation in react-native-keychain, where the build output is going to `./desktop/node_modules/react-native-keychain/desktop/qtkeychain/build/lib/libqt5keychain.a`, but the `status-react` project is looking for it in `node_modules/react-native-keychain/desktop/qtkeychain/build/lib/x86_64-linux-gnu/libqt5keychain.a`. I see that it is coming from https://github.com/status-im/react-native-keychain/blob/master/desktop/CMakeLists.txt#L11 but have no idea why it is only picking it up for the consumer side and not the producer side.

One workaround that works is the one proposed in the PR, where we change the whole `CMAKE_INSTALL_PREFIX` instead of just `qtkeychain_STATIC_LIB`.


mandrigin:
@PombeirP it looks like you wanted to merge it to `status-im:master` and not upstream.

PR #211, opened 2019-05-03
Fix SecurityLevel constants.

cknitt:


mandrigin:
@cknitt thanks a lot!

mandrigin:
@oblador I've merged this change. Maybe it would be nice to release a new version containing it, but I don't yet know how...

PR #224, opened 2019-07-02
Add a section for how to unbreak Jest test compile

genuinezack:
Describe how to mock the library so that Jest tests will successfully compile for apps that depend on the library upstream.


genuinezack:
👍 thanks for the edits and merge

PR #240, opened 2019-08-22
Update getSupportedBiometryType policy

petekp:
Updates the policy used for `canEvaluatePolicy` to `LAPolicyDeviceOwnerAuthentication` which enables developers to determine the device's `biometryType` even if the permissions are denied or blocked.

### DIFF
--- a/RNKeychainManager/RNKeychainManager.m
+++ b/RNKeychainManager/RNKeychainManager.m
@@ -270,7 +270,7 @@ RCT_EXPORT_METHOD(getSupportedBiometryType:(RCTPromiseResolveBlock)resolve rejec
 {
   NSError *aerr = nil;
   LAContext *context = [LAContext new];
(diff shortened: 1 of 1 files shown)


RNKeychainManager/RNKeychainManager.m
```diff
@@ -270,7 +270,7 @@ - (OSStatus)deleteCredentialsForServer:(NSString *)server
 {
   NSError *aerr = nil;
   LAContext *context = [LAContext new];
-  BOOL canBeProtected = [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:&aerr];
+  BOOL canBeProtected = [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthentication error:&aerr];
```
petekp: This policy requires iOS 9. Since this library is targeting iOS 9, this shouldn't need any OS version checks.

PR #261, opened 2019-10-31
Support macOS Catalyst

leonid-shevtsov:
Great news! I'm successfully using this package in my Catalyst project.

However, the [`SecRequestSharedWebCredential`](https://developer.apple.com/documentation/security/1617896-secrequestsharedwebcredential
) and `SecAddSharedWebCredential` functions are not supported in Catalyst, so the corresponding wrapper methods need to be conditionally excluded.


oblador:
Thanks!

PR #292, opened 2020-01-07
Update getSupportedBiometryType docs

petekp:
Removes a notice under `getSupportedBiometryType` that is no longer valid after a recent API change (https://github.com/oblador/react-native-keychain/pull/240) . This method always returns the biometry type, if the device supports it. It only returns `null` if the device doesn't support any of the biometry types.


oblador:
Thanks!

PR #302, opened 2020-03-05
Add support for customizing Android biometric prompt

jenskuhrjorgensen:
Add four string props to Android biometric prompt:
`promptInfoTitle`
`promptInfoSubtitle`
`promptInfoDescription`
`promptInfoNegativeBtnText`

There are a few more [props](https://developer.android.com/reference/androidx/biometric/BiometricPrompt.PromptInfo.Builder#public-methods_1) that could be added and used in the BiometricPrompt, but I will leave that to another PR (if someone needs it).

### DIFF
--- a/KeychainExample/App.js
+++ b/KeychainExample/App.js
@@ -88,7 +88,8 @@ export default class KeychainExample extends Component {
       const options = {
         authenticationPrompt: "Authentification needed",
         promptInfoTitle: "Authentication needed",
-        promptInfoSubtitle: "Some descriptive subtitle",
+        promptInfoSubtitle: "Subtitle",
+        promptInfoDescription: "Some descriptive text",
         promptInfoNegativeBtnText: "Cancel",
       }
       const credentials = await Keychain.getGenericPassword(options);
--- a/KeychainExample/App.js
+++ b/KeychainExample/App.js
@@ -13,7 +13,7 @@ import SegmentedControlTab from 'react-native-segmented-control-tab';
 import * as Keychain from 'react-native-keychain';
 
 const ACCESS_CONTROL_OPTIONS = ['None', 'Passcode', 'Password'];
-const ACCESS_CONTROL_OPTIONS_ANDROID = ['Default'];
+const ACCESS_CONTROL_OPTIONS_ANDROID = ['None'];
 const ACCESS_CONTROL_MAP = [
   null,
   Keychain.ACCESS_CONTROL.DEVICE_PASSCODE,
--- a/KeychainExample/App.js
+++ b/KeychainExample/App.js
@@ -86,14 +86,13 @@ export default class KeychainExample extends Component {
   async load() {
     try {
       const options = {
-        authenticationPrompt: "Authentication needed",
-        promptInfoOptions: {
-          title: "Authentication needed",
-          subtitle: "Subtitle",
-          description: "Some descriptive text",
-          negativeBtnText: "Cancel",
+        authenticationPrompt: {
+          title: 'Authentication needed',
+          subtitle: 'Subtitle',
+          description: 'Some descriptive text',
+          negativeBtnText: 'Cancel',
         },
-      }
+      };
       const credentials = await Keychain.getGenericPassword(options);
       if (credentials) {
         this.setState({ ...credentials, status: 'Credentials loaded!' });
@@ -123,7 +122,7 @@ export default class KeychainExample extends Component {
       const reply = await Keychain.setSharedWebCredentials(
         'server',
         'username',
-        'password',
+        'password'
       );
       console.log(`setSharedWebCredentials: ${JSON.stringify(reply)}`);
     } catch (err) {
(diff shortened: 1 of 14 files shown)


typings/react-native-keychain.d.ts
```diff
@@ -72,6 +72,10 @@ declare module 'react-native-keychain' {
     securityLevel?: SECURITY_LEVEL;
     storage?: STORAGE_TYPE;
     rules?: SECURITY_RULES;
+    promptInfoTitle: string;
```
oblador: I don't think it makes sense to change the API to be required. For example one might only use this on iOS. I'd change it so that the current values are default with an option to override.

oblador:
Even better would be to use the strings from this closed PR (that were never used) as default: https://github.com/cladjules/react-native-keychain/tree/4beddc67541c2d1e8c1641f54714a0ebb1676556/android/src/main/res

oblador:
Better source is this one: https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-biometric-release/biometric/res/

However I guess this means we'd need to amend the license to be MIT + Apache

typings/react-native-keychain.d.ts
jenskuhrjorgensen: Didn't think about that. I've made biometric prompt title and negative button optional again.

jenskuhrjorgensen:
@oblador 

> Even better would be to use the strings from this closed PR (that were never used) as default: https://github.com/cladjules/react-native-keychain/tree/4beddc67541c2d1e8c1641f54714a0ebb1676556/android/src/main/res

I was also looking at [that ](https://github.com/oblador/react-native-keychain/issues/304), but I think this PR can do without it for now. Could be awesome to have in another release though! 😃

vascofg:
I'd like to ask to at least expose these strings as constants (the ones from `androidx.biometric:biometric`). They are quite useful for e.g. showing Toasts / Alerts to the user when something goes wrong

jenskuhrjorgensen:
@vascofg Shouldn't this be handled by the biometric prompt itself? I think that is the intention with the "hint" text under the fingerprint icon (and it should be translated according to the language selected in the OS):
Screenshot_20200309-140418
Screenshot_20200309-140408

vascofg:
Indeed, but in case you manually trigger the prompt afterwards I'd like to
display a relevant error message to the user, e.g. when they lockout
biometrics (too many wrong attempts).

jenskuhrjorgensen:
Aah, that would make sense. However, I don't have the time to put it in this PR. It would be awesome if you could create PR for that! 

@oblador I don't have the time to put any other features into this PR, but I think your suggestions are well-suited for another PR as well. This PR is the minimum required to make it possible to customize the biometric prompt on Android the same way it can be done on iOS.

typings/react-native-keychain.d.ts
```diff
@@ -72,6 +72,10 @@ declare module 'react-native-keychain' {
     securityLevel?: SECURITY_LEVEL;
     storage?: STORAGE_TYPE;
     rules?: SECURITY_RULES;
+    promptInfoTitle?: string;
```
vonovak: I'd really find it preferable for the options to live in their own object and not mix in with the other options. ie.

```
promptOptions: {
  title?: string,
  subtitle?: string,
  // ... etc
}
```

jenskuhrjorgensen: I though about doing that, but I didn't now what style you wanted, so I just mixed them with the existing props. I'll fix this tomorrow.

jenskuhrjorgensen: @vonovak Done

index.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import { NativeModules, Platform } from 'react-native';
+import {PromptInfoOptions} from "react-native-keychain"
```
dattran-cxa, line 3: Hi @jenskuhrjorgensen , please check it, why do we import `PromptInfoOptions` from  `react-native-keychain`

jenskuhrjorgensen: I have no idea. Fixed :)

oblador:
@jenskuhrjorgensen & @vonovak: The current names and the duplication between iOS and Android didn't make sense to me, so I unified them under `authenticationPrompt`. Please have a look and see what you think.

KeychainExample/App.js
```diff
@@ -85,7 +85,15 @@ export default class KeychainExample extends Component {
 
   async load() {
     try {
-      const credentials = await Keychain.getGenericPassword();
+      const options = {
+        authenticationPrompt: {
+          title: 'Authentication needed',
+          subtitle: 'Subtitle',
+          description: 'Some descriptive text',
+          negativeBtnText: 'Cancel',
```
hbelen, line 93: This key should be `cancel` instead of `negativeBtnText `.

hbelen, line 93:
```suggestion
          cancel: 'Cancel',
```

lucaashrq:
When this will be released?

oblador:
It's already out in 6.0.0 :-)

Issue #306, opened 2020-03-05
RNKeychainManager.hasInternetCredentialsForServer was called with 2 arguments but expects 1 argument


vascofg:
Getting this error on iOS with version 5.0.0 of the library.
It seems this method was updated from 4.0.5 on the Javascript side but not on the iOS side.

vascofg:
@oblador @OleksandrKucherenko PR created

etaiklein:
I received something similar to this: https://github.com/oblador/react-native-keychain/issues/212
Could there be a similar solution as well?

vascofg:
Don't think so @etaiklein, since that call on Javascript actually matches the number of arguments on the iOS code:
```js
return RNKeychainManager.setGenericPasswordForOptions(
    options,
    username,
    password
  );
```

matches this one (plus promise handlers)

```objective-c
RCT_EXPORT_METHOD(setGenericPasswordForOptions:(NSDictionary *)options
                  withUsername:(NSString *)username
                  withPassword:(NSString *)password
                  resolver:(RCTPromiseResolveBlock)resolve
                  rejecter:(RCTPromiseRejectBlock)reject)
```

oblador:
Fixed in [5.0.1](https://github.com/oblador/react-native-keychain/releases/tag/v5.0.1).

cladjules:
Seems like it's broken again in 6.0.0

PR #307, opened 2020-03-05
Fix broken hasInternetCredentials on iOS

vascofg:
On the latest release (5.0.0) `hasInternetCredentials` function is broken on iOS, because Javascript passes some extra options (as described on #306).
This PR adds this extra parameter.


oblador:
Thanks! Fixed in 5.0.1 by removing the passed option in JS instead :-)

Issue #305, opened 2020-03-05
Improve error handling on getInternetCredentials exception - Android


vascofg:
Whenever getInternetCredentials throws an exception, it's currently really cumbersome to get the code thrown by Biometric Manager.
The only way I could find was to parse `err.message`, which has a format like `code: xx, message: xpto`.

We should expose the error in a better way, and also be able to differentiate between actual Errors decrypting credentials and simply scanning a not enrolled fingerprint (which in my Galaxy S8 is causing an error with just the message `[Error: Authentication failed. User Not recognized.]`.

vascofg:
Failing on `onAuthenticationFailed` is causing another issue.
If you scan an incorrect fingerprint, the process fails but the prompt stays open.
If you then scan a correct one, it closes but nothing happens

PR #312, opened 2020-03-10
Don't fail authentication flow onAuthenticationFailed

vascofg:
As described in #305, failing the whole flow in the `onAuthenticationFailed` callback seems incorrect.
The BiometricPrompt stays open if your fingerprint is not recognized for some reason (a common occurence), but fails to complete after matching.

This PR disables onAuthenticationFailed entirely.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -813,8 +813,6 @@ public class KeychainModule extends ReactContextBaseJavaModule {
     @Override
     public void onAuthenticationFailed() {
       final CryptoFailedException error = new CryptoFailedException("Authentication failed. User Not recognized.");
-
-      onDecrypt(null, error);
     }
 
     /** trigger interactive authentication. */
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -809,12 +809,6 @@ public class KeychainModule extends ReactContextBaseJavaModule {
       }
     }
 
-    /** Called when a biometric is valid but not recognized. */
-    @Override
-    public void onAuthenticationFailed() {
-      final CryptoFailedException error = new CryptoFailedException("Authentication failed. User Not recognized.");
-    }
-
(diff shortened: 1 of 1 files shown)


simlrh:
Tested this locally and `getGenericPassword` now works as I would expect it to, only rejecting when "Too many attempts" is reached, and resolving if the fingerprint is recognised after an initial failure.

vonovak:
hello @vascofg and thanks for the PR. Can you please add a little more context that would help me evaluate what is done here?  You mention #305 - I'm assuming you're referring to the second comment

> Failing on onAuthenticationFailed is causing another issue.
If you scan an incorrect fingerprint, the process fails but the prompt stays open.
If you then scan a correct one, it closes but nothing happens

You're saying
> If you then scan a correct one, it closes but nothing happens

why is that the case?

> This PR disables onAuthenticationFailed entirely.

Why is that the correct solution to the problem? Do I understand it correctly that when you scan a wrong fingerprint (or mis-scan a correct one), previously this would reject the promise, and now it won't?
Thank you!

vascofg:
Your assumption is correct. I believe this is the correct solution for the problem as I can't see many use cases where you want to fail the entire flow just because a user mis-scanned a fingerprint.

The BiometricPrompt takes care of retrying and actually failing with an error `onAuthenticationError` if you e.g. scan the wrong fingerprint too many times

vonovak:
@vascofg ok, to clarify 

> If you scan an incorrect fingerprint, the process fails but the prompt stays open.
If you then scan a correct one, it closes but nothing happens

do I understand it that this means

1. call getGenericPassword()
2. scan incorrect fingerprint, promise rejects
3. scan a correct one, it closes but nothing happens (=promise getGenericPassword() already rejected so nothing can happen)

?

thanks

vascofg:
Yes that is correct

Vojtech Novak <notifications@github.com> escreveu no dia quinta, 2/04/2020
à(s) 17:02:

vonovak:
okay, thanks, this makes sense. If someone is interested in knowing that `onAuthenticationFailed` happened, you'll need to use events for that https://reactnative.dev/docs/communication-ios#calling-react-native-functions-from-native-events

PR #326, opened 2020-04-04
Restore default behaviour for setGenericPassword on Android

hbelen:
**PR Description**

The default behaviour for `setGenericPassword` is different for iOS vs Android in v5.0.0 of this library. This PR aims to fix that.

**Details**

In v4.0.5 of `react-native-keychain` the default implementation allows users to set and get generic passwords in both iOS and Android without having to authenticate themselves. By passing `options` it is possible to modify this default behaviour.

In v5.0.0 the above still holds true for iOS. However, the default behaviour changed on Android devices that support biometrics. Such devices will prompt users to authenticate themselves even when passwords are stored without passing an `accessControl` value as an option, since they use the best available secured storage by default when setting passwords.

**Fix**

This PR restores the default behaviour on Android. Settings and getting secrets no longer forces a user to authenticate themselves by default. If passwords need to be stored in a more secure way, the `accessControl` option can be used.

**Recordings**

I have made some videos to illustrate the difference in behaviour. The file linked below contains recordings for using `getGenericPassword` and `setGenericPassword` with default settings in the following environments:
- v4.0.5 on Android
- master on iOS
- master on Android
- master + changes from this PR on Android

[react-native-keychain-recordings.zip](https://github.com/oblador/react-native-keychain/files/4432490/react-native-keychain-recordings.zip)

### DIFF
--- a/KeychainExample/App.js
+++ b/KeychainExample/App.js
@@ -13,7 +13,7 @@ import SegmentedControlTab from 'react-native-segmented-control-tab';
 import * as Keychain from 'react-native-keychain';
 
 const ACCESS_CONTROL_OPTIONS = ['None', 'Passcode', 'Password'];
-const ACCESS_CONTROL_OPTIONS_ANDROID = ['Default'];
+const ACCESS_CONTROL_OPTIONS_ANDROID = ['None'];
 const ACCESS_CONTROL_MAP = [
   null,
   Keychain.ACCESS_CONTROL.DEVICE_PASSCODE,
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Get security level that is supported on the current device with the current OS. 
 
 | Key                        | Platform     | Description                                                                                      | Default                                                      |
 | -------------------------- | ------------ | ------------------------------------------------------------------------------------------------ | ------------------------------------------------------------ |
-| **`accessControl`**        | All          | This dictates how a keychain item may be used, see possible values in `Keychain.ACCESS_CONTROL`. | _None_ (iOS), `BIOMETRY_ANY` default for Android.            |
+| **`accessControl`**        | All          | This dictates how a keychain item may be used, see possible values in `Keychain.ACCESS_CONTROL`. | _None_                                                       |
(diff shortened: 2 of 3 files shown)


KeychainExample/App.js
```diff
@@ -13,7 +13,7 @@ import SegmentedControlTab from 'react-native-segmented-control-tab';
 import * as Keychain from 'react-native-keychain';
 
 const ACCESS_CONTROL_OPTIONS = ['None', 'Passcode', 'Password'];
-const ACCESS_CONTROL_OPTIONS_ANDROID = ['Default'];
```
oblador, line 16: I guess we want to have both `None` and `Fingerprint` here?

README.md
```diff
@@ -181,7 +181,7 @@ Get security level that is supported on the current device with the current OS.
 
 > Note #1: `BIOMETRY_ANY`, `BIOMETRY_CURRENT_SET`, `BIOMETRY_ANY_OR_DEVICE_PASSCODE`, `BIOMETRY_CURRENT_SET_OR_DEVICE_PASSCODE` - recognized by Android as a requirement for Biometric enabled storage (Till we got a better implementation);
 >
-> Note #2: For Android we support only two states: `Default` (use the best available secured storage) and `Fingerprint` (use only biometric protected storage);
+> Note #2: For Android we support only two states: `None` (default) and `Fingerprint` (use only biometric protected storage);
```
oblador, line 184: I'm not sure, but I think we should be having a third state that reflects on the previous `Default` behavior.

oblador: Which I think would be analogous to `BIOMETRY_ANY` for iOS

KeychainExample/App.js
hbelen: That option is added on the fly for devices that support it.

https://github.com/oblador/react-native-keychain/blob/2d9ade4097bc8aba63d8f67c75e95456140f4e7f/KeychainExample/App.js#L189

README.md
oblador: I wasn't thinking right here, this is correct 👍

PR #333, opened 2020-04-22
Adds missing interface for promptInfoOptions attribute

koyta:
Adds a missing (for some reason) interface. Due to its absence, compilation errors occurred without the skipLibCheck rule enabled to `true`.

Also, as I see, you have an existing `AuthenticationPrompt` interface, but with optional `title`. This one requires `title`.


oblador:
Fixed in https://github.com/oblador/react-native-keychain/commit/cb41f7835cf8e6108704bc645e1845001946f32b

PR #348, opened 2020-05-28
Branches/filter slow devices

john-y-pazekha:
# The problem
The performance on _certain_ devices (very small number) is unacceptably slow, which makes this issue block the release of the app. 

Two issues are open at the moment: https://github.com/oblador/react-native-keychain/issues/314 and https://github.com/oblador/react-native-keychain/issues/337

# Investigation
The investigation is in progress. So far the findings are:
The performance degradation happened after the Ciphers have been refactored in the latest version. The refactoring was combined with adding new functionality (biometry). At the moment it is not clear which of the two factors is responsible.

All three Ciphers currently used underwent the refactoring changes. One of them, namely RSA cipher, is confirmed to not have any problems; the other two are unsure.

Reportedly, those devices which experienced bad performance were back to normal after the library was downgraded to the pre-refactoring version.

# This PR
The purpose of this PR is not to fix the issue but to unblock the release of the app. This gives the developers room to investigate and fix properly.

## The approach
The idea is to minimise the impact of changes introduced during the refactoring. Once the app detects that it is running on the affected device, the compromised ciphers will not be loaded. This may degrade the cryptographic capability, though. Therefore this solution cannot be seen as permanent.

## What and where
* The patch is found in `KeychainModule.java`. See the comment near it.
* Detection of the affected devices is done by `Issue314Filter.java` and `Issue337Filter.java`. Once the issues are fixed, the respective filter must be deleted. Filters are already marked as deprecated.
* For easy injecting the filters and warm-up wrapper, a builder `KeychainModuleBuilder ` was introduced.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -39,9 +39,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
-
-import javax.crypto.Cipher;
 
 @SuppressWarnings({"unused", "WeakerAccess", "SameParameterValue"})
 public class KeychainModule extends ReactContextBaseJavaModule {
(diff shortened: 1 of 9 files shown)


android/src/main/java/com/oblador/keychain/KeychainModule.java
```diff
@@ -133,10 +131,16 @@
...
+  public KeychainModule(@NonNull final ReactApplicationContext reactContext, @NonNull final IDeviceFilter deviceFilter) {
     super(reactContext);
     prefsStorage = new PrefsStorage(reactContext);
 
+  // Exclude the refactored code from the execution path to keep it aligned with the old version as much as possible.
+  // NOTE: This may result in reduction of the encryption capabilities.
+  // TODO Unwrap this workaround once the problem is resolved
+  if (deviceFilter.isDeviceAffected()) {
+    // TODO Add a fallback Cipher
+  } else {
```
john-y-pazekha, line 143: This is the actual workaround; the rest is just preparation and harness around it.

The fallback cipher is TBD.

Issue #383, opened 2020-08-25
getGenericCredentials never resolves or rejects due to ShortBufferException in Google Pixel 4


HuiSF:
**Description**

When use `setGenericCredentials` to write a long content (string length > 2400) as `password` into key store (AESCBC) in Google **Pixel 4**, `getGenericCredentials` will never resolve, nor ever reject.

**Details**

react-native-keychain version `6.1.1`.

When `getGenericCredentials` hangs, below Java exception was logged

```
08-21 20:54:31.465 22572 25907 W CipherStorageBase: ShortBufferException is not expected
08-21 20:54:31.465 22572 25907 W CipherStorageBase: java.lang.IllegalStateException: ShortBufferException is not expected
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at javax.crypto.CipherInputStream.getMoreData(CipherInputStream.java:136)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at javax.crypto.CipherInputStream.read(CipherInputStream.java:249)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at javax.crypto.CipherInputStream.read(CipherInputStream.java:225)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at com.oblador.keychain.cipherStorage.CipherStorageBase.copy(CipherStorageBase.java:487)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at com.oblador.keychain.cipherStorage.CipherStorageBase.decryptBytes(CipherStorageBase.java:364)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at com.oblador.keychain.cipherStorage.CipherStorageKeystoreAesCbc.decryptBytes(CipherStorageKeystoreAesCbc.java:233)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at com.oblador.keychain.cipherStorage.CipherStorageKeystoreAesCbc.decrypt(CipherStorageKeystoreAesCbc.java:142)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at com.oblador.keychain.cipherStorage.CipherStorageKeystoreAesCbc.decrypt(CipherStorageKeystoreAesCbc.java:160)
08-21 20:54:31.465 22572 25907 W CipherStorageBase: 	at com.oblador.keychain.KeychainModule.decryptToResult(KeychainModule.java:623)
[...]

PR #260, opened 2019-10-21
AndroidX biometric manager

OleksandrKucherenko:
implementation of Biometric with fingerprint scan interaction.

Major changes:
- reused concept of https://github.com/oblador/react-native-keychain/pull/195
- Added unit tests
- keychain is a major class for attaching UI interactions. Storage ciphers are low-level classes and should not work with UI
- made base cipher class that simplifies the implementation of other storages
- optimized yarn.lock
- added postinstall action for the sample that attaches the latest source of the library and clean up the folder after that
- upgraded gradle to 5.6.2 version
- upgraded RN to 0.61 version
- added re-packaging code for facebook.conceal library (latest version has problems https://github.com/facebook/conceal/issues/199)
- used AndroidX biometric library
- tested with api19, api21, api23, api28, real device with fingerprints enabled
- added ability to force usage of specific cipher storage
- refactored code for future upgrades (passcode, password, pin, etc.)
- updated documentation, api, ts & js code
- added several specific for Android flags (build scripts)
- migrated to latest gradle + build plugin, gradle kotlin (*.kts) scripts
- example configured as includedBuild (gradle feature)
- added developer-contributor documentation (in details shows how to setup Android Studio)
- tested on emulator and devices, compatibility with prev version of stored data added
- tested upgrade and downgrade secured storage scenarios (documented in readme file)
- added "warming up" thread that helps java to load all crypto library code in memory and make library works much much faster... instead of 2-12 seconds to access stored secret we spent now 10-30ms. 

image

PR #385, opened 2020-08-27
Reintroduce a fix for strongbox decryption data corruption (#383)

cshfang:
A PR (https://github.com/oblador/react-native-keychain/pull/288) a while back was merged to fix an issue where decrypting the keychain using a `CipherInputStream` causes certain Google Pixel devices to throw an exception and hang.

The change seems to have been inadvertently reverted in the large refactoring that took place in https://github.com/oblador/react-native-keychain/pull/260.

This PR reintroduces that fix.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
@@ -31,7 +31,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.crypto.Cipher;
-import javax.crypto.CipherInputStream;
 import javax.crypto.CipherOutputStream;
 import javax.crypto.NoSuchPaddingException;
 import javax.crypto.spec.IvParameterSpec;
@@ -351,7 +350,7 @@ abstract public class CipherStorageBase implements CipherStorage {
     throws GeneralSecurityException, IOException {
     final Cipher cipher = getCachedInstance();
 
-    // decrypt the bytes using a CipherInputStream
+    // decrypt the bytes using cipher.doFinal()
     try (ByteArrayInputStream in = new ByteArrayInputStream(bytes);
          ByteArrayOutputStream output = new ByteArrayOutputStream()) {
 
@@ -360,11 +359,8 @@ abstract public class CipherStorageBase implements CipherStorage {
         handler.initialize(cipher, key, in);
       }
 
-      try (CipherInputStream decrypt = new CipherInputStream(in, cipher)) {
-        copy(decrypt, output);
-      }
-
-      return new String(output.toByteArray(), UTF8);
+      byte[] decryptedBytes = cipher.doFinal(bytes, IV.IV_LENGTH, bytes.length - IV.IV_LENGTH);
+      return new String(decryptedBytes, UTF8);
     } catch (Throwable fail) {
       Log.w(LOG_TAG, fail.getMessage(), fail);
 
@@ -470,23 +466,6 @@ abstract public class CipherStorageBase implements CipherStorage {
   public static String getDefaultAliasIfEmpty(@Nullable final String service, @NonNull final String fallback) {
     //noinspection ConstantConditions
     return TextUtils.isEmpty(service) ? fallback : service;
-  }
-
-  /**
-   * Copy input stream to output.
-   *
-   * @param in  instance of input stream.
-   * @param out instance of output stream.
-   * @throws IOException read/write operation failure.
-   */
-  public static void copy(@NonNull final InputStream in, @NonNull final OutputStream out) throws IOException {
-    // Transfer bytes from in to out
-    final byte[] buf = new byte[BUFFER_READ_WRITE_SIZE];
-    int len;
-
-    while ((len = in.read(buf)) > 0) {
-      out.write(buf, 0, len);
-    }
   }
   //endregion
 
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
@@ -498,7 +498,7 @@ abstract public class CipherStorageBase implements CipherStorage {
       cipher.init(Cipher.ENCRYPT_MODE, key);
     };
 
-    public static final DecryptBytesHandler<InputStream> decrypt = (cipher, key, input) -> {
+    public static final DecryptBytesHandler decrypt = (cipher, key, input) -> {
       cipher.init(Cipher.DECRYPT_MODE, key);
     };
   }
@@ -515,8 +515,8 @@ abstract public class CipherStorageBase implements CipherStorage {
       final byte[] iv = cipher.getIV();
       output.write(iv, 0, iv.length);
     };
-    /** Read initialization vector from input bytes array and configure cipher by it. */
-    public static final DecryptBytesHandler<byte[]> decrypt = (cipher, key, input) -> {
+    /** Read initialization vector from input stream and configure cipher by it. */
+    public static final DecryptBytesHandler decrypt = (cipher, key, input) -> {
       final IvParameterSpec iv = readIv(input);
       cipher.init(Cipher.DECRYPT_MODE, key, iv);
     };
@@ -553,9 +553,9 @@ abstract public class CipherStorageBase implements CipherStorage {
       throws GeneralSecurityException, IOException;
   }
 
-  /** Handler for configuring cipher by initialization data from input. */
-  public interface DecryptBytesHandler<T> {
-    void initialize(@NonNull final Cipher cipher, @NonNull final Key key, @NonNull final T input)
+  /** Handler for configuring cipher by initialization data from input stream. */
+  public interface DecryptBytesHandler {
+    void initialize(@NonNull final Cipher cipher, @NonNull final Key key, @NonNull final InputStream input)
       throws GeneralSecurityException, IOException;
   }
 
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
@@ -14,6 +14,7 @@ import com.oblador.keychain.SecurityLevel;
 import com.oblador.keychain.exceptions.CryptoFailedException;
 import com.oblador.keychain.exceptions.KeyStoreAccessException;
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
```diff
@@ -360,11 +359,8 @@ protected String decryptBytes(@NonNull final Key key, @NonNull final byte[] byte
         handler.initialize(cipher, key, in);
```
aeirola, line 359:
```suggestion
        handler.initialize(cipher, key, bytes);
```

Now that the whole decrypt process works with byte arrays, it might be better to change the `DecryptBytesHandler` to take the input as a byte array, instead of an input stream.

This change also requires the `DecryptBytesHandler` interface input parameter to be changed to `byte[]`, but shouldn't require other changes as the parameter is only passed to overloaded methods which already handle byte arrays as well as InputStreams.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
```diff
@@ -351,7 +350,7 @@ protected String decryptBytes(@NonNull final Key key, @NonNull final byte[] byte
     throws GeneralSecurityException, IOException {
     final Cipher cipher = getCachedInstance();
 
-    // decrypt the bytes using a CipherInputStream
+    // decrypt the bytes using cipher.doFinal()
```
aeirola, line 353: Might be good to add a comment in the code with a link to the issue, or this PR, describing why we don't use input streams. Otherwise someone might be eager to refactor this again, without testing on a pixel device first :)

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
```diff
@@ -351,7 +350,7 @@ protected String decryptBytes(@NonNull final Key key, @NonNull final byte[] byte
...
     try (ByteArrayInputStream in = new ByteArrayInputStream(bytes);
          ByteArrayOutputStream output = new ByteArrayOutputStream()) {
```
aeirola, line 355:
```suggestion
    try {
```

I don't think the input stream and output stream are really used anymore, since the decrypt operation now works with byte arrays. These should probably be removed.

aeirola (changes requested):
In general, the code looks good! Would still like to clean it up a bit by removing the mostly unused byte array streams. Otherwise I'm worried they might cause some confusion in the future.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
cshfang: No problem at, makes complete sense to me.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
cshfang: 👍 This was a miss. I should've cleaned this up.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
cshfang: 👍 This was a miss. I should've cleaned this up.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
```diff
@@ -505,7 +504,7 @@ public static String getDefaultAliasIfEmpty(@Nullable final String service, @Non
     public static IvParameterSpec readIv(@NonNull final byte[] bytes) throws IOException {
       final byte[] iv = new byte[IV_LENGTH];
 
-      if (IV_LENGTH <= bytes.length)
```
cshfang, line 529: When switching over to this overloaded method handling `byte[]`, this condition was triggered. Reading through this method, I believe the intent may have been to validate that the input bytes array is actually *greater* than `IV_LENGTH`.

aeirola: Whoa, strange bug. Wonder if this code was ever used before.

aeirola:
The code looks good, and normal encoding and decoding seems to work OK on Android. But when trying to get stored values which are protected with fingerprint authentication, I get the a `javax.crypto.IllegalBlockSizeException` error caused by `android.security.KeyStoreException: Invalid input length`. Apparently the AES and RSA cipher storages use the same `decryptBytes` method, and the RSA cipher used for fingerprints doesn't allow decyrpting data using `doFinal` with arbitrary length input.

Might be that we need to move the modified `decryptBytes` method from `CipherStorageBase` to the `CipherStorageKeystoreAesCbc` class, so that we don't interfere with the workings of the other cipher storage implementations.

This issue wasn't present in the previous fix for the same issue, as the Android fingerprints weren't supported yet at the time.

cshfang:
> The code looks good, and normal encoding and decoding seems to work OK on Android. But when trying to get stored values which are protected with fingerprint authentication, I get the a `javax.crypto.IllegalBlockSizeException` error caused by `android.security.KeyStoreException: Invalid input length`. Apparently the AES and RSA cipher storages use the same `decryptBytes` method, and the RSA cipher used for fingerprints doesn't allow decyrpting data using `doFinal` with arbitrary length input.
> 
> Might be that we need to move the modified `decryptBytes` method from `CipherStorageBase` to the `CipherStorageKeystoreAesCbc` class, so that we don't interfere with the workings of the other cipher storage implementations.
> 
> This issue wasn't present in the previous fix for the same issue, as the Android fingerprints weren't supported yet at the time.

Ah thanks for the good catch. Based on what you're saying, it makes sense to me to try to make the modified `decryptBytes` method the exception for AES rather than the norm. I'll work on this today.

cshfang:
@aeirola I pushed a refactor to override `decryptBytes` within the `CipherStorageKeystoreAesCbc` class. It seems like the changes are still holding up in my testing but I only tested with a fairly contrived RSA set/get. I'm not sure how to replicate the `android.security.KeyStoreException: Invalid input length` you found above so may have to lean on you for further testing.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
```diff
@@ -553,9 +553,9 @@ void initialize(@NonNull final Cipher cipher, @NonNull final Key key, @NonNull f
       throws GeneralSecurityException, IOException;
   }
 
-  /** Handler for configuring cipher by initialization data from input stream. */
-  public interface DecryptBytesHandler {
-    void initialize(@NonNull final Cipher cipher, @NonNull final Key key, @NonNull final InputStream input)
+  /** Handler for configuring cipher by initialization data from input. */
+  public interface DecryptBytesHandler<T> {
```
aeirola, line 557: I'm wondering if making this interface generic causes some confusion, especially when the different `CipherStorage` implementations don't explicitly state which type of `DecryptBytesHandler` they expect. Maybe would be better to not touch these common files too much, and just omit using the `handler` in the overridden `CipherStorageKeystoreAesCbc.decryptBytes`.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
```diff
@@ -215,6 +218,32 @@ protected Key generateKey(@NonNull final KeyGenParameterSpec spec) throws Genera
...
+    final Cipher cipher = getCachedInstance();
+
+    // decrypt the bytes using cipher.doFinal(). Using a CipherInputStream for decryption has historically led to issues
+    // on the Pixel family of devices.
+    // see https://github.com/oblador/react-native-keychain/issues/383
+    try {
+      // read the initialization vector from bytes array
+      if (null != handler) {
+        handler.initialize(cipher, key, bytes);
+      }
```
aeirola, line 237:
```suggestion
      final IvParameterSpec iv = IV.readIv(bytes);
      cipher.init(Cipher.DECRYPT_MODE, key, iv);
```

In order to not have to make the handler interface generic.

I realise that this will make the implementation deviate a bit from the other `CipherStorage` classes, but probably best to keep this workaround as contained as possible.

aeirola (approved):
Code looks good, and works well in my testing. Might personally prefer to keep the changes even more local to the `CipherStorageKeystoreAesCbc` class by not using generics for the `DecryptBytesHandler` interface, but this is more of a matter of taste. Feel free to change it or ignore my comments :)

Before merging I'll try to contact people who've faced the issue and see wether they would be willing to help out testing this PR with actual Titan M devices.

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.java
cshfang: Sure, I understand the concern. I considered just adding comments to specify the generics handled but didn't want to pigeonhole the interface too much. Also opted to not stray from the original implementation too much and therefore kept the `handler` but your suggestion makes sense also and I'm happy to make that change.

cshfang:
> Before merging I'll try to contact people who've faced the issue and see wether they would be willing to help out testing this PR with actual Titan M devices.

FWIW, I have been testing this against my physical Pixel 4 device as well, which I believe utilizes the Titan M.

aeirola:
@Metroidaron @nicolashemonic Hi, I remember that you've previously helped this project by testing issues on Pixel devices in #208 . Now we're in a situation where we have faced a regression in the support for Pixel devices with Titan M security chips. In order to verify that our fix is valid, I'd like to ask wether you would be able to help out once again by checking if the latest code in this PR still works on your devices and use cases. The code from this PR can be installed with `npm install github:cshfang/react-native-keychain#fix-strongbox-decryption`. Thanks again!

aeirola:
Unfortunately there wasn't much interest in testing the fix on additional devices. Would have been nice with some extra validations with different devices, but won't let that block the deployment of the fix as the code looks good and you already tested on one Pixel 4 device. I'll merge this and see if we can get it released soon.

aeirola:
@oblador Do we have a schedule for the next release? Releaseing this fix would be useful for a certain group of Android users.

oblador:
@aeirola It's released in 6.2.0!

nicolashemonic:
Hi @aeirola @oblador 

Sorry for the silence because I was in long vacation 🏖 and busy when back to the office.

I upgraded my app to `react-native-keychain@6.2.0` and I confirm that the fix works well on Pixel 3a Android 11. The authentication (very long token stored in keychain) is not lost when the app is closed and opened.

Thanks !

PR #397, opened 2020-09-17
fix: Xcode 12 compatibility

radko93:
Latest Xcode 12 fails to build without a module to depend on `React-Core` directly hence this change is necessary for all native modules on iOS. This change requires React Native 0.60.2 or newer. For more details please check: https://github.com/facebook/react-native/issues/29633#issuecomment-694187116


Dirklectisch:
Well it took me a couple hours of fiddling before I found this PR that fixed it. Thanks 👍

oblador (approved):

Issue #318, opened 2020-03-24
Android: Wrapped error: User not authenticated


alphec:
I get this error when trying to retrieve credentials stored in the keychain on Android devices. IMHO this is the effect of following exception being thrown: [UserNotAuthenticatedException](https://developer.android.com/reference/android/security/keystore/UserNotAuthenticatedException).

This is my code:
```
  useEffect(() => {
    if (!loginStarted) {
      return;
    }

    Keychain.getSupportedBiometryType()
      .then(data => {
        console.log('Supported biometry: ' + data);
        setBiometry(!!data);
        if (!!data) {
          // Try auto login
          Keychain.getGenericPassword({service: service})
            .then(credentials => {
              if (!credentials) {
                // no credentials stored, continue with normal login
                console.log('No saved credentials.');
                return;
              }

              // auto login with stored credentials
              if (credentials.username && credentials.password) {
                console.log('Trying auto login with saved credentials: ' + JSON.stringify(credentials));
                login(credentials.username, credentials.password);
              }
            })
            .catch(e => console.log('Error getting credentials: ' + e));
        }
      })
      .catch(e => console.log('Error getting biometry: ' + e));
  }, [loginStarted]); 
```

The error is thrown at `Keychain.getGenericPassword`. On iOS the code works fine. Any idea or workaround for this?

skicson:
I noticed similar behavior where the `onAuthenticationSucceeded` was being called (in `KeychainModule.java` but I was still getting the UserNotAuthenticatedException.   Basically the decryption key's authentication was expiring before the cipher could get initialized.  (probably since I was running in a debug environment, but could also happen with slower devices)

I was able to get it to work by extending the `UserAuthenticationValidityDuration` attribute of the key from 1 second to 5.  In my case, it was using the RSA key, so the change was in the `CipherStorageKeystoreRsaEcb.java` file, around line 228
```
    return new KeyGenParameterSpec.Builder(alias, purposes)
      .setBlockModes(BLOCK_MODE_ECB)
      .setEncryptionPaddings(PADDING_PKCS1)
      .setRandomizedEncryptionRequired(true)
      .setUserAuthenticationRequired(true)
      .setUserAuthenticationValidityDurationSeconds(5)
      .setKeySize(ENCRYPTION_KEY_SIZE);
```

hoangcxa:
I'm having the same problem with android Samsung s9. 
I'm using this method to save credentials.
` setInternetCredentials(server, username, password, {
[...]

PR #399, opened 2020-09-22
Fix of issue 318

Zo2m4bie:
Fix issue 318 with FaceId for android
There're two issues here. First of all Pixel 4 doesn't work. Fix with setUserAuthenticationValidityDurationSeconds allow Pixel to work however it asks face not all the time, which sometime looks strange when users expect to see face recognition but they don't see it .
Second problem is Samsung. Samsung face recognition isn't secure that's why you see "User not authenticated" issue.
I rework logic with cipher and now it works for Samsung(Samsung doesn't ask face) and for Pixel.
Also RSA were removed as it doesn't work with a new solution


andrejcesen:
@Zo2m4bie Hey, could you elaborate on why RSA doesn't work with your solution?

Zo2m4bie:
@andrejcesen well I just wasn't able to manage RSA cipher to work with BiometricPrompt. I tried to do it with RSA at the beginning but I faced with error when I passed cipher into BiometricPrompt. I guess it's related to pair of public and private keys that are used in RSA. As when cipher is passed to the BiometricPrompt, it changes with some biometric data and we need to manage vector bytes from this cipher to restore cipher for decryption. As I understood this logic doesn't work fully with RSA.
With RSA works .setUserAuthenticationRequired(true). To unlock it you need to trigger BiometricPrompt that ask you you fingerprint or face id. In case if phone vendor consider that specific biometric auth method insecure, next time you'll try to access cipher you will receive "User not authorised" error. In my solution, because cipher is passed into BiometricPrompt, the phone suggest to use only secure variants of biometric and unsecure variants are hidden.

andrejcesen:
Awesome, thanks for sharing @Zo2m4bie.

CWolfs:
I'm not sure removing RSA is the right approach here to fix this issue. The fixes seem very good but this would break backwards compatibility for anyone who requires RSA set up and aren't easily able to migrate over to AES.

Is there a more backwards compatible approach you can take?

Zo2m4bie:
@CWolfs hm, very good point. For now I see only one available solution is to create separate migration method that will load data from RSA and move it into AES. The only problem of this solution that it will ask fingerprint twice. First one to load data from RSA storage and second one to save it into a new storage

andrejcesen:
@CWolfs @Zo2m4bie That's exactly what I did, based off of your solution :) I'll try to publish a PR with the changes here in the upcoming days.

AdriaRios:
Hi guys!

I tried this branch @Zo2m4bie on a Pixel 4. 

There is an issue, when I call to `setGenericPassword` the prompt is presented when it shouldn't. Any idea?

Zo2m4bie:
Hello @AdriaRios It's the expected behavior. You can look at this prompt as an entry point to your Cipher. You need to pass it to encrypt or decrypt data.

AdriaRios:
Oh really? Is because the main branch doesn't work like that, and it's a bit weird the behaviour

CWolfs:
@AdriaRios If you're using RSA on the main branch then it's asymmetric crypto (public/private keys). So to encrypts the data using the public key - so you don't need to authenticate. When you go to decrypt you use the private key and so to access that you need to authenticate.

I haven't looked at @Zo2m4bie's PR properly but I assume since he's using only AES, which is symmetric, then you need access to the KeyStore for both encrypting and decrypting and so requires you to unlock the Keystore with an authentication for boths times.

That's my assumption on the last bit anyway.

Zo2m4bie:
@CWolfs you are absolutely right

Zo2m4bie:
@oblador my case with Samsung device was fixed. However let it be open until resolving a new issue described in 318 thread because this solution works for all cases

UgurGumushan:
After using this fork for react-native-keychain and making adjustments (adding authenticationPrompt on setInternetCredentials), instead of having https://github.com/oblador/react-native-keychain/issues/318, now I have "Could not encrypt data with alias"
The implementation which works on iOS (real device) wouldn't work on Android emulator.

YanislavRemitly:
> I believe this issue should have been resolved as of [7.0.0](https://github.com/oblador/react-native-keychain/releases/tag/v7.0.0). Can anyone confirm?

Still have this issue on version 8.1.1

DorianMazur:
Closing this PR as it's really old

Issue #279, opened 2019-12-06
getSupportedBiometryType() always returns the result


webforuse:
getSupportedBiometryType() always returns the result
even if Faceid or Touchid is not Enrolled
although according to the documentation should return null

petekp:
@webforuse The `getSupportedBiometryType` API recently changed and this method only returns `null` if no biometry type is supported, otherwise it always returns the device's biometry type. I submitted a PR to update the docs: https://github.com/oblador/react-native-keychain/pull/292

ePoromaa:
Any suggestions on how to check if the device is enrolled for any biometrics?

edersonmberti:
I noticed that this behavior occurs only on iOS devices. 

The solution I found was change de policy _LAPolicyDeviceOwnerAuthentication_ to _LAPolicyDeviceOwnerAuthenticationWithBiometrics_ inside node_modules folder on file "RNKeychainManager.m", to be more precise on line 273. 

``BOOL canBeProtected = [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:&aerr];``

After that, I used [patch-package](https://www.npmjs.com/package/patch-package) to generate a patch to change every time that packages are installed.

``npx patch-package react-native-keychain``

I don't know if it's the best solution, but works.

pasine:
I read a potential workaround on iOS is to use `canImplyAuthentication` with `Keychain.ACCESS_CONTROL.BIOMETRY_CURRENT_SET`, but I don't understand how this would work with Android, as `canImplyAuthentication` works on iOS only.

jess-dovetail:
@edersonmberti I wonder if you can submit a PR of your change? Perhaps the authors and main collaborators can review it and merge it in.

PR #428, opened 2021-01-28
Fixed iOS returning biometry type even when not enrolled.

jesuscc1993:
Makes the following statement from the readme become true:

> ### `getSupportedBiometryType()`
> 
> Get what type of hardware biometry support the device has. Resolves to a `Keychain.BIOMETRY_TYPE` value when supported, otherwise `null`.
> 
> > This method returns `null`, if the device haven't enrolled into fingerprint/FaceId. Even though it has hardware for it.

See issue [#279](https://github.com/oblador/react-native-keychain/issues/279) for more details.

Thanks to [edersonmberti](https://github.com/edersonmberti) for [his fix](https://github.com/oblador/react-native-keychain/issues/279#issuecomment-656734014).

### DIFF
--- a/RNKeychainManager/RNKeychainManager.m
+++ b/RNKeychainManager/RNKeychainManager.m
@@ -298,7 +298,7 @@ RCT_EXPORT_METHOD(getSupportedBiometryType:(RCTPromiseResolveBlock)resolve
 {
   NSError *aerr = nil;
   LAContext *context = [LAContext new];
-  BOOL canBeProtected = [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthentication error:&aerr];
+  BOOL canBeProtected = [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:&aerr];
 
   if (!aerr && canBeProtected) {
     if (@available(iOS 11, *)) {


oblador:
This was the initial implementation, I don't think it makes sense to change this back and forth. Maybe better to expose another API that checks this? See https://github.com/oblador/react-native-keychain/pull/240

M-Jas:
@oblador This is an extremely useful thing to check for numerous reasons. A quick example is preventing the FaceId Scan run if the user has turned it off in the device settings the next time they try to log in.

This is your package and I'm very thankful for all of your hard work, but if you aren't going to accept these changes what's the point of keeping the availability of checking for if the device isn't enrolled into fingerprint/FaceId in the docs? [#240](https://github.com/oblador/react-native-keychain/pull/240) was merged in 2 years ago.

343max:
We had to fork react native keychain because we needed this fix. As far as I can see there is no other way for the app to know if the user has enabled FaceID/TouchID. Also the documentation is clearly wrong as it says it will return null if the user has not enrolled for Biometrics while it returns it even if the user has only set up a passcode.

oblador:
Since behavior is changing with #449 I decided to merge this one too to keep them in line. You didn't allow me to rebase the branch, so I cherry-picked it to master instead: https://github.com/oblador/react-native-keychain/commit/a89b7d0bfb070587d839ad8900a91eea1ae3021e

jesuscc1993:
> You didn't allow me to rebase the branch
I don't understand this sentence, was there something I did wrong with the PR setup?
Speaking of which, should I close it now that you cherry-picked the commit?
@oblador

oblador:
Yep, published in [7.0.0](https://github.com/oblador/react-native-keychain/releases/tag/v7.0.0).

PR #447, opened 2021-03-14
Fix biometric retry on OnePlus devices

### LINKED ISSUE #335: App hangs on OnePlus 6T and above!!
Hello,
I have recently used this package to implement biometric authentication in my react-native app. It works perfectly, but I am facing an issue in OnePlus 6T and above devices.

Below are the listed issues that i am facing.
1.  Biometrics works perfectly if i authenticate successfully. But if i authenticate with wrong finger, the prompt goes away and my app hangs. I have to restart the app to make it working again.
2. I cannot cancel the biometrics prompt, as i only see a fingerprint icon on the phone. No cancel button like other android devices to cancel it.

To be honest i am just a react-native developer and not a native android or IOS developer (I am currently learning to become a native developer). So i have less clue of what's wrong. But i did my research and found this https://forums.oneplus.com/threads/problems-about-oneplus-6t-fingerprint-api.944959/. I would try to fix this on my own but it would take me ages. Please help.

### LINKED ISSUE #446: Biometric retry does not work on OnePlus devices
Due to the broken implementation of `BiometricPrompt` in OxygenOS (Android OS used on OnePlus devices), `onAuthenticationFailed` callback is leaving `BiometricPrompt` in a dismissed but orphaned state. This means that biometric retry will not be presented to the user, even though no error is being thrown. Even after manually retrying the `BiometricPrompt.authenticate()`, the prompt is not shown, due to its orphaned state.

**Details and investigation [here](https://forums.oneplus.com/threads/oneplus-7-pro-fingerprint-biometricprompt-does-not-show.1035821/#post-21710422)**

All OnePlus models after 6 (6T, 7, 7 Pro, 7T, etc) are affected by this. I can reproduce it on my OnePlus 6T - details below.

*Reproduction steps (Using example app in this repo)*
1. Type `test` in both `username` and `password`
2. Select the following options: `AccessControl: Fingerprint`, `SecurityLevel: Hardware`, `Storage: RSA`
3. Press `Save` and wait for it to be saved
4. Press load and observe the biometric sensor activating
5. Scan non-enrolled finger
6. Observe that the biometric prompt is gone (no retry) and the app is in a non-interactable state
7. Put app into background (or open multitasking)
8. Bring the app back to the foreground and observe that it is back to normal state
9. Press `Load` and scan enrolled finger to see that everything works

*My device*
Screenshot_20210313-182449

oneplus_biometric_retry_issue

I've found the solution using parts of the suggestion in the forum post linked above. Will send PR in a bit.

sgal:
Fixes #446 
Fixes #335 
This change addresses the issue with biometric retry on OnePlus devices. The main approach is taken from the incredible investigation done by the user in [this forum post](https://forums.oneplus.com/threads/oneplus-7-pro-fingerprint-biometricprompt-does-not-show.1035821/#post-21710422).

The idea is to call `BiometricPrompt.cancelAuthentication()` after each `onAuthenticationFailed` received. After manual cleanup, the biometric auth is retried (without creating a new `waitForResult` block). When the user runs out of retries the flow completes gradually in `onAuthenticationError` which throws `CryptoFailedException`.

I verified this solution on my OnePlus 6T running OxygenOS 10.

**TODO**
- [x] Add unit tests

## Retrying until sensor is disabled (too many attempts)
fixed_retry_too_many_attempts

## Retrying once with following success
fixed_retry_success

### DIFF
--- a/android/src/main/java/com/oblador/keychain/decryptionHandler/DecryptionResultHandlerInteractiveBiometricManualRetry.java
+++ b/android/src/main/java/com/oblador/keychain/decryptionHandler/DecryptionResultHandlerInteractiveBiometricManualRetry.java
@@ -14,11 +14,10 @@ public class DecryptionResultHandlerInteractiveBiometricManualRetry extends Decr
   private BiometricPrompt presentedPrompt;
   private Boolean didFailBiometric = false;
 
-  public DecryptionResultHandlerInteractiveBiometricManualRetry(@NonNull FragmentActivity activity,
-                                                                @NonNull ReactApplicationContext reactContext,
+  public DecryptionResultHandlerInteractiveBiometricManualRetry(@NonNull ReactApplicationContext reactContext,
                                                                 @NonNull CipherStorage storage,
                                                                 @NonNull BiometricPrompt.PromptInfo promptInfo) {
-    super(activity, reactContext, storage, promptInfo);
+    super(reactContext, storage, promptInfo);
   }
 
   /** Manually cancel current (invisible) authentication to clear the fragment. */
@@ -72,6 +71,8 @@ public class DecryptionResultHandlerInteractiveBiometricManualRetry extends Decr
   /** trigger interactive authentication. */
   @Override
   public void startAuthentication() {
+    FragmentActivity activity = getCurrentActivity();
+
     // code can be executed only from MAIN thread
     if (Thread.currentThread() != Looper.getMainLooper().getThread()) {
       activity.runOnUiThread(this::startAuthentication);
@@ -79,12 +80,15 @@ public class DecryptionResultHandlerInteractiveBiometricManualRetry extends Decr
       return;
     }
 
-    this.presentedPrompt = authenticateWithPrompt();
+    this.presentedPrompt = authenticateWithPrompt(activity);
   }
 
   /** trigger interactive authentication without invoking another waitResult() */
   private void retryAuthentication() {
     Log.d(LOG_TAG, "Retrying biometric authentication.");
+
+    FragmentActivity activity = getCurrentActivity();
+
     if (Thread.currentThread() != Looper.getMainLooper().getThread()) {
       try {
         /*
@@ -100,6 +104,6 @@ public class DecryptionResultHandlerInteractiveBiometricManualRetry extends Decr
       return;
     }
 
-    this.presentedPrompt = authenticateWithPrompt();
+    this.presentedPrompt = authenticateWithPrompt(activity);
   }
 }
--- a/android/src/main/java/com/oblador/keychain/decryptionHandler/DecryptionResultHandlerInteractiveBiometricManualRetry.java
+++ b/android/src/main/java/com/oblador/keychain/decryptionHandler/DecryptionResultHandlerInteractiveBiometricManualRetry.java
@@ -1,0 +1,105 @@
+package com.oblador.keychain.decryptionHandler;
+
+import android.os.Looper;
+import android.util.Log;
+
+import androidx.annotation.NonNull;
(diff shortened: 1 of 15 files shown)


build.gradle.kts
```diff
@@ -52,7 +52,7 @@ allprojects {
       }
 
       when ("${requested.group}:${requested.name}") {
-        "com.facebook.react:react-native" -> useVersion("0.61.2")
+        "com.facebook.react:react-native" -> useVersion("0.63.4")
```
sgal, line 55: This version is in `yarn.lock` and the build fails without this change.

android/src/main/java/com/oblador/keychain/decryptionHandler/DecryptionResultHandlerInteractiveBiometricManualRetry.java
```diff
@@ -0,0 +1,105 @@
...
+      return;
+    }
+
+    super.onAuthenticationError(errorCode, errString);
+  }
+
+  /** Called when a biometric (e.g. fingerprint, face, etc.) is presented but not recognized as belonging to the user. */
+  @Override
+  public void onAuthenticationFailed() {
+    Log.d(LOG_TAG, "Authentication failed: biometric not recognized. " + (presentedPrompt == null));
```
sgal, line 56:
```suggestion
    Log.d(LOG_TAG, "Authentication failed: biometric not recognized.");
```

android/src/test/java/com/oblador/keychain/KeychainModuleTests.java
```diff
@@ -150,6 +150,7 @@ public void testFingerprintConfigured_api23() throws Exception {
     //   Fingerprints are configured
     //   fingerprint feature is ignored by android os
     ReactApplicationContext context = getRNContext();
+    shadowOf(context.getPackageManager()).setSystemFeature(PackageManager.FEATURE_FINGERPRINT, true);
```
sgal, line 153: The test was failing without this line, which is present in another test. I assumed that it was a miss from some previous change and added this.

android/src/test/java/com/oblador/keychain/KeychainModuleTests.java
```diff
@@ -448,6 +450,11 @@ public void testDowngradeBiometricToAes_api28() throws Exception {
     final JavaOnlyMap options = new JavaOnlyMap();
     options.putString(Maps.SERVICE, "dummy");
 
+    final JavaOnlyMap promptOptions = new JavaOnlyMap();
+    promptOptions.putString(KeychainModule.AuthPromptOptions.TITLE, "Title");
```
sgal, line 454: This test was failing because `PromptInfo` requires title and negativeText to be set even when `setUserConfirmationRequired(false)`.

sgal:
I tested these changes on my OnePlus 6T with the latest OxygenOS available for that phone (Android 10). Would be great if someone could test it on a Samsung, or any other device that does not have a biometric retry bug.

sgal:
@oblador @vonovak Could you please have a look?

oblador (approved):
Awaiting confirmation on Samsung hardware 👍

sgal:
Verified on Samsung Galaxy S10 running Android 10 - works as expected even with retries.

vonovak (approved):
hi and thanks for the PR! Lots of changes so I only had a quick look and it seems fine to me! I love that you added some automated tests... too bad they do not run in CI but that's another story :)

oblador:
@vonovak I'll see what I can do about that 👍

oblador:
As promised: https://github.com/oblador/react-native-keychain/pull/453

PR #449, opened 2021-03-27
Add support for strong biometry

sgal:
`androidx.biometric:biometric:1.1.0` has added APIs that support detection and usage of strong biometry (Class 3). Support of such access control is useful for apps that require extra security, like banks, payments etc.

I successfully tested strong biometry use-cases on my OnePlus 6T, but still need to verify it on at least a Samsung device that has Class 2 (face recognition) and Class 3 (fingerprint) sensors. 

One important thing that I would like to verify is whether numerous reports of `User not authenticated` errors are related to the fact that only Class 3 biometric can unlock the Keystore. In the current implementation, this lib checks whether there is at least Class 2 biometric present and when calling `BiometricPrompt.authenticate` without `CryptoObject` param it does not unlock the Keystore, resulting in an endless loop of errors.

## Changes
- Bumped `androidx.biometrics` to `1.1.0` [Changelog](https://developer.android.com/jetpack/androidx/releases/biometric#1.1.0)
- Added `androidx.lifecycle:lifecycle-viewmodel` because activity parts were removed from `biometrics` lib.
- Added `androidx.fragment:fragment` because fragment parts were removed from `biometrics` lib.
- Removed activity from Manifest file, because it is no longer present.

## Notable changes from the update of `androidx.biometric` lib
- Pixel 4 biometric detection is fixed https://issuetracker.google.com/issues/170406186
- Smaller bundle size on older devices
- `BiometricPrompt` got the API to fallback to device credential if biometric sensor is locked (too many retries)


sgal:
I was able to confirm the fact that class 2 biometry does not unlock the key store. 

On Samsung S10 there exists a Class 2 face recognition and Class 3 Fingerprint. Both are offered in the biometric prompt, but only Fingerprint is able to authorize the decryption of the value in the key store. Face recognition gives `User not authenticated` error.

sgal:
I tested the update on Samsung Galaxy S20, please find the before and after pictures below. The change does force strong biometric by utilising a [fake crypto object trick](https://android.googlesource.com/platform/frameworks/support/+/c70e4a93bc85cc874a265a67336315991aff82e5/biometric/biometric/src/main/java/androidx/biometric/BiometricFragment.java#375).

| Before | After |
| --- | --- |
| before | after |

KeychainExample/App.js
```diff
@@ -87,6 +87,9 @@ export default class KeychainExample extends Component {
   async load() {
     try {
       const options = {
+        accessControl: this.state.accessControl,
```
sgal, line 90: These must be passed to `getGenericPassword` call to force strong biometrics.

README.md
```diff
@@ -151,12 +151,21 @@ Sets a shared web credential. Resolves to `true` when successful.
 
 Inquire if the type of local authentication policy is supported on this device with the device settings the user chose. Should be used in combination with `accessControl` option in the setter functions. Resolves to `true` if supported.
 
-### `getSupportedBiometryType()`
+### `getSupportedBiometryType()` _(deprecated)_
```
sgal, line 154: The reason this is deprecated is that it sometimes returns Face recognition as a supported biometric, even though it cannot be used for crypto operations.

sgal:
While debugging this, I found that having `.setUserAuthenticationRequired(true)` on the Key creation makes it protected by Class 3 biometric. So you cannot event call `cipher.init(Cipher.DECRYPT_MODE, key);` without authenticating with Class 3 biometric. Android OS does not make it very clear that you need to authenticate with Class 3 biometrics in order to unlock the key - this is where all `User not authenticated` errors are coming from.

Interesting thing here is that you can authenticate any Key that does not require a user authentication with Class 3 biometrics which in turn will unlock all of the Keys that do require user authentication. This is how the [fake crypto object trick](https://android.googlesource.com/platform/frameworks/support/+/c70e4a93bc85cc874a265a67336315991aff82e5/biometric/biometric/src/main/java/androidx/biometric/CryptoObjectUtils.java#240) is working inside `androidx.biometrics`. They generate AES key in memory and wrap it in `CryptoObject` that is being passed to `BiometricPrompt.authenticate` call, to force the usage of Class 3 biometrics. The generated AES key does not require user authentication to be used.

sgal:
@oblador @vonovak This is no longer a draft, so please review when you have time.

KeychainExample/App.js
```diff
@@ -87,6 +87,9 @@ export default class KeychainExample extends Component {
...
+        securityLevel: this.state.securityLevel,
```
oblador, line 91: These two can be omitted per our discussion

KeychainExample/App.js
oblador: As we discussed it would be preferable if this could be inferred like it works on iOS instead.

sgal: Inferred this from the storage type and removed the params.

sgal:
After further investigation and discussion with @oblador we decided to replace old flaky biometric detection and usage logic with the one that relies on strong biometrics. This means that this change might be breaking for some users, so new major release is required.

android/src/main/java/com/oblador/keychain/DeviceAvailability.java
```diff
@@ -53,13 +53,4 @@ public static boolean isPermissionsGranted(@NonNull final Context context) {
     // before api28
     return context.checkSelfPermission(Manifest.permission.USE_FINGERPRINT) == PERMISSION_GRANTED;
   }
-
-  public static boolean isDeviceSecure(@NonNull final Context context) {
```
sgal, line 57: This method wasn't used.

oblador (approved):

PR #457, opened 2021-04-27
Optimized getCipherStorageForCurrentAPILevel method

Vito96:
reordered the execution of the instructions to optimize the getCipherStorageForCurrentAPILevel method.

In the case of isBiometry = false it is useless to execute the getCapabilityLevel method which can be calculated in case the RSA private key is missing.

I realized this because not using the warm up, the first de-encryption was still very slow due to this particular.

Furthermore, this re-order does not cause problems for the warm up because it is valued there with isBiometry = true.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -728,19 +728,20 @@ public class KeychainModule extends ReactContextBaseJavaModule {
     for (CipherStorage variant : cipherStorageMap.values()) {
       Log.d(KEYCHAIN_MODULE, "Probe cipher storage: " + variant.getClass().getSimpleName());
 
+      // if biometric supported but not configured properly than skip
+      if (variant.isBiometrySupported() && !isBiometry) continue;
+
       // Is the cipherStorage supported on the current API level?
       final int minApiLevel = variant.getMinSupportedApiLevel();
-      final int capabilityLevel = variant.getCapabilityLevel();
       final boolean isSupportedApi = (minApiLevel <= currentApiLevel);
 
       // API not supported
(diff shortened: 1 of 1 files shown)


vonovak:
hello and thank you for the PR! Can you please take a look at the failing test and make it pass? Thank you! 💯

Vito96:
```
   Test testGetSecurityLevel_NoBiometry_NoSecuredHardware_api28 FAILED

   java.lang.AssertionError:
   Expected: not null
        but: was null
       at com.oblador.keychain.KeychainModuleTests.testGetSecurityLevel_NoBiometry_NoSecuredHardware_api28 (KeychainModuleTests.java:433)
       
```
       
 Why is this RSA test required?

vonovak:
>  Why is this RSA test required?

Hello, I'm on mobile so I don't see the details of the test. Maybe it's not neccessary any more, maybe the test should be modified - my point is, we cannot merge a PR if it makes tests fail. 

We need to have a clear understanding of whether your change introduced a bug or whether the failing test should be modified.

If you have time to claeify that, perfect. Otherwise I'll try to get to it at some point.

oblador:
This should be fixed by #505 AFAIK, can you can confirm @Vito96?

Vito96:
Not totally, this PR (#505) of course optimizes the RSA check for the initial test using another key dimension, also properly uses the short condition evaluation here
Schermata 2021-10-04 alle 10 37 30
:

But it is useless evaluate isBiometry in this point (row 753)
Schermata 2021-10-04 alle 10 38 49

This PR can optimize the reorder of this lines for not compute del capability level for variant which isBiometrySupported is true and isBiometry is false.

Anyway, with the #505 PR, now the tests should be fine.

PR #495, opened 2021-08-11
Avoid calling biometric APIs if not using biometrics (closes #494)

### LINKED ISSUE #494: should not call biometric APIs if not using biometrics
KeychainModule.java should avoid calling biometric APIs when the application does not plan to use biometric-based access control.  Specifically, probing is always done to determine if the various biometric features are available.  This can be avoided by making a one-line fix inside getCipherStorageForCurrentAPILevel().  I will create a pull request.

MarkCSmith:


oblador:
Thanks for your PR, this is fixed in #505

PR #497, opened 2021-08-20
feat: Allow to synchronize keychain items to iCloud

### LINKED ISSUE #156: Cannot see my keychain in iCloud
All seems to work as expected. My question is:
Is there a way to see the generated keychain, perhaps via Keychain Access.app or in iCloud?

Thanks for any suggestion.

m-ruhl:
Closes #156


RNKeychainManager/RNKeychainManager.m
```diff
@@ -248,23 +256,28 @@ - (void)insertKeychainEntry:(NSDictionary *)attributes
   }
 }
 
-- (OSStatus)deletePasswordsForService:(NSString *)service
+- (OSStatus)deletePasswordsForOptions:(NSDictionary *)options
```
oblador, line 259: Does it really make sense to have this as a part of the delete query?

m-ruhl: @oblador can we clarify this?

mrousavy: I think Joel meant if we need the `synchronized` flag for the reset functions, because it doesn't really change behaviour - does it?

m-ruhl: From my understanding, with the `sychronized` flag the entry is deleted from the iCloud keychain, otherwise in the device keychain

NagaiKoki:
this pr is still working?

bumpingChris:
@m-ruhl hi, thanks for doing this. Quick question - does this Synchronized flag work for Android - to persist the entries in keychain beyond a single device?

jesus-castro3:
> @m-ruhl hi, thanks for doing this. Quick question - does this Synchronized flag work for Android - to persist the entries in keychain beyond a single device?

Going by the docs and changes made by @m-runl,  its iOS only

pleerock:
seems good one to have completed and merged

PR #505, opened 2021-09-17
Speed optimisations for Android

bjornallvin:
This is a set of changes to Android native component to optimise speed. 

The main reason for slowness is that RSA is a very cpu intensive algorithm. The changes in this PR try to mitigate this by removing unneeded calls and I've also added a less secure (and thereby much faster) encryption option, to be used for testing if the cipher works on the device. 

General policy changes is that the option to upgrade ciphers now defaults to false. And upgrades will only happen for keys stored with FacebookConseal cipher.

**KeychainModule.java - _getGenericPassword_**
We no longer always check for available ciphers on the device when getting a password. We instead only do this test if we should upgrade ciphers. 

**CipherStorageKeystoreRsaEcb.java**
Adding an alternative key size to use for testing the cipher. Real encryption uses 3072 bits, but for testing purpose we now only use 512 bits. Which is a lot faster to calculate. 

**CipherStorageBase.java  - _getCapabilityLevel_**
When evaluating which ciphers exists and at what capability level. The check to see if secure hardware is available was removed as this is a very expensive operation and it's not needed for evaluation purpose in that step.

**CipherStorageBase.java - _getCipherStorageForCurrentAPILevel_**
Small optimisation of order of evaluation

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -299,8 +299,7 @@ public class KeychainModule extends ReactContextBaseJavaModule {
         final String accessControl = getAccessControlOrDefault(options);
         final boolean useBiometry = getUseBiometry(accessControl);
         cipher = getCipherStorageForCurrentAPILevel(useBiometry);
-
-      }else{
+      } else {
         cipher = getCipherStorageByName(storageName);
       }
 
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -286,20 +286,31 @@ public class KeychainModule extends ReactContextBaseJavaModule {
         return;
       }
 
-      // get the best storage
-      final String accessControl = getAccessControlOrDefault(options);
-      final boolean useBiometry = getUseBiometry(accessControl);
-      final CipherStorage current = getCipherStorageForCurrentAPILevel(useBiometry);
+      final String storageName = resultSet.cipherStorageName;
       final String rules = getSecurityRulesOrDefault(options);
-
       final PromptInfo promptInfo = getPromptInfo(options);
-      final DecryptionResult decryptionResult = decryptCredentials(alias, current, resultSet, rules, promptInfo);
+
+      CipherStorage cipher = null;
+
+      // Only check for upgradable ciphers for FacebookConseal as that
+      // is the only cipher that can be upgraded
+      if (rules.equals(Rules.AUTOMATIC_UPGRADE) && storageName.equals(KnownCiphers.FB)){
+        // get the best storage
+        final String accessControl = getAccessControlOrDefault(options);
+        final boolean useBiometry = getUseBiometry(accessControl);
+        cipher = getCipherStorageForCurrentAPILevel(useBiometry);
+
+      }else{
+        cipher = getCipherStorageByName(storageName);
+      }
+
+      final DecryptionResult decryptionResult = decryptCredentials(alias, cipher, resultSet, rules, promptInfo);
 
       final WritableMap credentials = Arguments.createMap();
       credentials.putString(Maps.SERVICE, alias);
       credentials.putString(Maps.USERNAME, decryptionResult.username);
       credentials.putString(Maps.PASSWORD, decryptionResult.password);
-      credentials.putString(Maps.STORAGE, current.getCipherStorageName());
+      credentials.putString(Maps.STORAGE, cipher.getCipherStorageName());
 
       promise.resolve(credentials);
     } catch (KeyStoreAccessException e) {
@@ -488,11 +499,11 @@ public class KeychainModule extends ReactContextBaseJavaModule {
     return getAliasOrDefault(service);
   }
 
-  /** Get automatic secret manipulation rules, default: Automatic Upgrade. */
+  /** Get automatic secret manipulation rules, default: No upgrade. */
   @Rules
   @NonNull
   private static String getSecurityRulesOrDefault(@Nullable final ReadableMap options) {
-    return getSecurityRulesOrDefault(options, Rules.AUTOMATIC_UPGRADE);
+    return getSecurityRulesOrDefault(options, Rules.NONE);
   }
 
   /** Get automatic secret manipulation rules. */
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/oblador/keychain/KeychainModule.java
```diff
@@ -488,11 +499,11 @@ private static String getServiceOrDefault(@Nullable final ReadableMap options) {
     return getAliasOrDefault(service);
   }
 
-  /** Get automatic secret manipulation rules, default: Automatic Upgrade. */
+  /** Get automatic secret manipulation rules, default: No upgrade. */
   @Rules
   @NonNull
   private static String getSecurityRulesOrDefault(@Nullable final ReadableMap options) {
-    return getSecurityRulesOrDefault(options, Rules.AUTOMATIC_UPGRADE);
+    return getSecurityRulesOrDefault(options, Rules.NONE);
```
oblador, line 506: Need to change README to reflect this.

android/src/main/java/com/oblador/keychain/KeychainModule.java
```diff
@@ -286,20 +286,31 @@ protected void getGenericPassword(@NonNull final String alias,
...
+
+      // Only check for upgradable ciphers for FacebookConseal as that
+      // is the only cipher that can be upgraded
+      if (rules.equals(Rules.AUTOMATIC_UPGRADE) && storageName.equals(KnownCiphers.FB)){
+        // get the best storage
+        final String accessControl = getAccessControlOrDefault(options);
+        final boolean useBiometry = getUseBiometry(accessControl);
+        cipher = getCipherStorageForCurrentAPILevel(useBiometry);
+
+      }else{
```
oblador, line 303:
```suggestion
      } else {
```

android/src/main/java/com/oblador/keychain/KeychainModule.java
```diff
@@ -286,20 +286,31 @@ protected void getGenericPassword(@NonNull final String alias,
...
+      CipherStorage cipher = null;
+
+      // Only check for upgradable ciphers for FacebookConseal as that
+      // is the only cipher that can be upgraded
+      if (rules.equals(Rules.AUTOMATIC_UPGRADE) && storageName.equals(KnownCiphers.FB)){
+        // get the best storage
+        final String accessControl = getAccessControlOrDefault(options);
+        final boolean useBiometry = getUseBiometry(accessControl);
+        cipher = getCipherStorageForCurrentAPILevel(useBiometry);
+
```
oblador, line 302:
```suggestion
```

android/src/main/java/com/oblador/keychain/KeychainModule.java
```diff
@@ -722,7 +733,7 @@ protected DecryptionResultHandler getInteractiveHandler(@NonNull final CipherSto
   /* package */ CipherStorage getCipherStorageForCurrentAPILevel(final boolean useBiometry)
     throws CryptoFailedException {
     final int currentApiLevel = Build.VERSION.SDK_INT;
-    final boolean isBiometry = (isFingerprintAuthAvailable() || isFaceAuthAvailable() || isIrisAuthAvailable()) && useBiometry;
+    final boolean isBiometry = useBiometry && (isFingerprintAuthAvailable() || isFaceAuthAvailable() || isIrisAuthAvailable());
```
oblador, line 736: 👌

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageFacebookConceal.java
```diff
@@ -152,6 +152,14 @@ public void removeKey(@NonNull final String alias) {
...
   }
 
+  @NonNull
+  @Override
+  protected KeyGenParameterSpec.Builder getKeyGenSpecBuilder(@NonNull final String alias, @NonNull final boolean isForTesting)
+    throws GeneralSecurityException {
+    throw new CryptoFailedException("Not designed for a call");
+  }
+
+
```
oblador, line 162:
```suggestion
```

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
```diff
@@ -173,10 +173,18 @@ public void decrypt(@NonNull final DecryptionResultHandler handler,
...
   //region Implementation
 
+  /** Get builder for encryption and decryption operations with required user Authentication. */
+  @NonNull
+  @Override
+  protected KeyGenParameterSpec.Builder getKeyGenSpecBuilder(@NonNull final String alias) throws GeneralSecurityException {
+    return getKeyGenSpecBuilder(alias, false);
+  }
+
+
```
oblador, line 183:
```suggestion
```

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreRsaEcb.java
```diff
@@ -209,25 +211,40 @@ private EncryptionResult innerEncryptedCredentials(@NonNull final String alias,
       this);
   }
 
+
```
oblador, line 214:
```suggestion
```

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreRsaEcb.java
```diff
@@ -209,25 +211,40 @@ private EncryptionResult innerEncryptedCredentials(@NonNull final String alias,
...
   /** Get builder for encryption and decryption operations with required user Authentication. */
   @NonNull
   @Override
   @SuppressLint("NewApi")
-  protected KeyGenParameterSpec.Builder getKeyGenSpecBuilder(@NonNull final String alias)
+  protected KeyGenParameterSpec.Builder getKeyGenSpecBuilder(@NonNull final String alias) throws GeneralSecurityException{
+    return getKeyGenSpecBuilder(alias, false);
+  }
+
+
```
oblador, line 223:
```suggestion
```

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreRsaEcb.java
```diff
@@ -209,25 +211,40 @@ private EncryptionResult innerEncryptedCredentials(@NonNull final String alias,
...
+  @SuppressLint("NewApi")
+  protected KeyGenParameterSpec.Builder getKeyGenSpecBuilder(@NonNull final String alias, @NonNull final boolean isForTesting)
     throws GeneralSecurityException {
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.M) {
       throw new KeyStoreAccessException("Unsupported API" + Build.VERSION.SDK_INT + " version detected.");
     }
 
     final int purposes = KeyProperties.PURPOSE_DECRYPT | KeyProperties.PURPOSE_ENCRYPT;
 
+    int keySize = ENCRYPTION_KEY_SIZE;
```
oblador, line 236: Ternary operator is cleaner I think

oblador (commented):
<3

bjornallvin:
I'm currently checking why one of the tests are failing :-)

android/src/main/java/com/oblador/keychain/KeychainModule.java
bjornallvin: Right! I'll get on that.

bjornallvin:
Tests are passing now. Just awaiting reviews.

oblador (approved):

hsjoberg:
Thanks, it appears that this PR finally fixed slow startup issues that I've been seeing on OnePlus 6 and possibly other devices (kept me stuck on version 4.0.5). 🚀

PR #512, opened 2021-10-03
Implement AES GCM for KeychainModule

bobowinca:
According to security articles during my research, AES GCM No-padding is more secure than the AES CBC.


gabrielmoncea:
@bobowinca what happened to this issue? Was it implemented in another PR?

bobowinca:
> 

I think it didn't get reviewed for a while, so I closed it.

PR #513, opened 2021-10-03
Implement AES GCM for KeychainModule

bobowinca:
According to the security articles during my research, AES GCM No-padding is more secure than the AES CBC.


DorianMazur:
AES GCM will be added with #678

PR #555, opened 2022-07-06
ios 13 check

rf1804:
'kSecUseDataProtectionKeychain' is only available on iOS 13.0 or newer

Enclose 'kSecUseDataProtectionKeychain' in an @available check to silence this warning

Probable Cause https://github.com/oblador/react-native-keychain/pull/536

### DIFF
--- a/RNKeychainManager/RNKeychainManager.m
+++ b/RNKeychainManager/RNKeychainManager.m
@@ -203,12 +203,8 @@ SecAccessControlCreateFlags accessControlValue(NSDictionary *options)
 
   NSMutableDictionary *mAttributes = attributes.mutableCopy;
 
-  if (@available(macOS 10.15, *)) {
-      if (@available(iOS 13.0, *)) {
-          mAttributes[(__bridge NSString *)kSecUseDataProtectionKeychain] = @(YES);
-      } else {
-          // Fallback on earlier versions
-      }
+  if (@available(macOS 10.15, iOS 13.0, *)) {
+      mAttributes[(__bridge NSString *)kSecUseDataProtectionKeychain] = @(YES);
   }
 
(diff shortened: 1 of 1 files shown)


oblador:
@hsjoberg Wdyt?

oblador:
I don’t really see why this check would even be neccessary as it shouldn’t enter this if startment at all on iOS. The nested if should also negate each other so that it would never enter the inner one unless i misunderstand how they work.

hsjoberg:
I'm a bit confused on how it would ever even enter that if-clause under iOS env.
I suppose we should go with the check considering the urgency, but do we know if macOS still enters everything correctly considering the irrationality of checking both macOS and iOS versions?

hsjoberg:
Okay the asterisk means that the code will run for every other platform...

image

https://developer.apple.com/forums/thread/88869

Suggestion:
```diff
-if (@available(macOS 10.15, *)) {
+if (@available(macOS 10.15, iOS 13.0)) {
```

rf1804:
> Okay the asterisk means that the code will run for every other platform...
> 
> image
> 
> https://developer.apple.com/forums/thread/88869
> 
> Suggestion:
> 
> ```diff
> -if (@available(macOS 10.15, *)) {
> +if (@available(macOS 10.15, iOS 13.0)) {
> ```

Your mentioned suggestion was giving error, correct implementation is

```diff
-if (@available(macOS 10.15, *)) {
+if (@available(macOS 10.15, iOS 13.0, *)) {
```
Same has been updated in PR, and tested on ios 12.4 and 15.5

@oblador @hsjoberg

rf1804:
@oblador @hsjoberg Can you please take a look.

oblador (approved):

Issue #525, opened 2022-01-20
getGenericPassword on Android sometimes fails and crashes app


jm90m:
getGenericPassword on android sometimes fails. I can't figure out why, it just crashes the app.

kriit24:
in my android it does not do anything code stopes on "await Keychain.getGenericPassword" and thats it

sgal:
@jm90m @kriit24 Could you folks please share more info on your environments?
Android version, Keychain options that you pass to the `getGenericPassword` would help investigating the issue.

hazim-j:
@jm90m @kriit24 Did you manage to find a fix for this?

Getting a similar experience as well but can't seem to figure out what exactly is causing it or how to reproduce. 

Only hacky workaround that I could find is if I run async code before calling `getGenericPassword()` that forces a state update + some delay. Otherwise app freezes until I eventually get a "App isn't responding" alert.

Example:
```js
setHack(true);
await delay(500);

const credentials = await Keychain.getGenericPassword();
```

For more context:
- Android version `12`
-  react-native-keychain version `8.1.1`
- Not passing any options to `getGenericPassword`

Lmk if there's any other information I can give to debug this.

kabirbaidhya:
I've dug deeper into this issue and have been able to consistently reproduce it and I've narrowed down when it happens.

**TLDR: It seems to occur when biometrics is triggered in the middle of unfinished UI navigation or rendering;
which blocks the main thread and makes the app unresponsive and eventually it crashes.** 

### Reproduction steps:
In an Android device try these:
1. Invoke `Keychain.getGenericPassword`
2. Without awaiting the promise, close the current screen and immediately navigate to a new screen that takes some time to render i.e. contains some UI operation. 
3. When the Biometrics prompt shows up, you'll notice in the overlay that the ongoing rendering in the background is already frozen. 
4. Either fail or pass the Biometrics. Both will crash the app. 

### Why? 

Looking through the android device logs, I've seen this occur only in cases where biometrics is triggered in the middle of navigation/rendering or when a UI operation is in-progress but not complete, and then the biometrics prompt is presented just in time.

[...]

PR #576, opened 2022-11-13
[Fix][Android] Avoid blocking threads to prevent deadlocks

frw:
Similar to https://github.com/oblador/react-native-keychain/pull/547, I've run into issues with main thread deadlocking when using this library to show a BiometricPrompt and have seen an increased rate of ANRs on Crashlytics.
Upon further digging it was due to `react-native-reanimated` attempting to acquire a lock on the main thread, but since this library waits on the JS thread for the BiometricPrompt to show and finish on the main thread, a deadlock occurs.
This library avoids blocking the JS thread by using a listener to resolve a promise when decryption finishes. Since there are no additional dependencies, this should be compatible up to API level 21.

This PR should fix https://github.com/oblador/react-native-keychain/issues/525

### DIFF
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
@@ -13,6 +13,7 @@ import androidx.annotation.Nullable;
 import com.oblador.keychain.KeychainModule.KnownCiphers;
 import com.oblador.keychain.SecurityLevel;
 import com.oblador.keychain.decryptionHandler.DecryptionResultHandler;
+import com.oblador.keychain.decryptionHandler.DecryptionResultListener;
 import com.oblador.keychain.exceptions.CryptoFailedException;
 import com.oblador.keychain.exceptions.KeyStoreAccessException;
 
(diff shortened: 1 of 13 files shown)


androideveloper:
For reanimated related issues, please refer to this PR and thread https://github.com/software-mansion/react-native-reanimated/pull/3440

frw:
Thanks @androideveloper for referring me to that thread! Glad to know something is also being done on the reanimated side to fix it.
However, I think we should still consider fixing this thread-blocking behavior as well since it is not ideal to be blocking a thread while waiting for the biometrics prompt to finish. Would appreciate any comments or feedback on the PR!

ca057:
I experienced the same issue and could resolve by using the version from this PR, all tests so far are successful. Is there anything I can support with to get this change / fix merged into upstream?

frw:
@oblador  Any chance you could take a look at this, seems like it's still affecting a few folks

kabirbaidhya:
Any progress / blockers on this PR? Can we prioritize this?

androideveloper:
Hi @frw! I'm going to have a look at your PRs in this repo. Did you already test these changes in a real app?

android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
```diff
@@ -127,11 +128,12 @@ public EncryptionResult encrypt(@NonNull final String alias,
   }
 
   @Override
-  @NonNull
-  public DecryptionResult decrypt(@NonNull final String alias,
-                                  @NonNull final byte[] username,
-                                  @NonNull final byte[] password,
-                                  @NonNull final SecurityLevel level)
+  public void decrypt(@NonNull final DecryptionResultHandler handler,
```
androideveloper, line 131: is handler not used anymore?

DorianMazur:
Fixed in 9.0.0

Issue #518, opened 2021-11-16
Android devices pass through biometric authentication even with passcode on unlock


MikaelCompile:
Hello!

So it seems that android devices pass through the authentication event from unlocking the phone. The issue right now is that we use biometrics to lock our app, but it doesn't seem to care wether we use fingerprint or device password to open the device, everything seems to just pass through and unlock our app as well. (provided the app is at least the first thing active when unlocking).

Is there any way to exert some control over this behaviour?

AlmiSelus:
Hi,

I've come across the same behaviour when using RSA based storage.

It seems that the reason (at least on the surface level) is the following line:
https://github.com/oblador/react-native-keychain/blob/6cafc99a26986772bc0fd30706db55877e103c3f/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreRsaEcb.java#L241

Is there any reason why the duration for user authentication validity is set strictly to 5?
If there is no need for this setting to be fixed, how about making it configurable from React API?

If you need, I can submit PR with my suggestion.

MikaelCompile:
Hi!
Thank you, this looks great! If nothing else, we can patch the library in our app.
It happens to be the same 5 seconds that i arrived to with my highly technical approach of counting in my heads each time before attempting it :)

Thanks again, this might be a life saver!

br. Mikael

MikaelCompile:
.setUserAuthenticationValidityDurationSeconds(5)  seems to deprecated in API 30, we ended up using the new method which takes two parameters: seconds and valid auth types.

We replaced the line with this: 
-      .setUserAuthenticationValidityDurationSeconds(5)
+      .setUserAuthenticationParameters(5, KeyProperties.AUTH_BIOMETRIC_STRONG)

and now it doesn't just go through biometrics with the screen unlock unless you opened the device with biometrics.

Vilinyexc:
The above mentioned solution didn't really last... Is there no way to disallow screen unlock to apply to the app? It doesn't seem to limit the options to biometrics at all, you can wizz past getGeneric or getInternetCredentials with whatever opened the phone and thats just unwanted behaviour. Scouring android resources and trying tweaks here and there and patching the library, but nothing seems to help. Does anyone have any sort of solution?

sgal:
@Vilinyexc There is no reliable way to prevent Keystore unlock after user is authenticated by strong biometry or device credential. It is how the Keystore is  implemented in Android.

[...]

PR #592, opened 2023-03-19
[Fix] [Android] Android devices pass through biometric authentication even with passcode on unlock (Issue #518)

nicolas-meilan:
- Always Ask access permissions on decrypt method instead of waiting an error.
- setUserAuthenticationValidityDurationSeconds is deprecated on API 30.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreRsaEcb.java
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreRsaEcb.java
@@ -125,18 +125,8 @@ public class CipherStorageKeystoreRsaEcb extends CipherStorageBase {
       // key is always NOT NULL otherwise GeneralSecurityException raised
       key = extractGeneratedKey(safeAlias, level, retries);
 
-      final DecryptionResult results = new DecryptionResult(
-        decryptBytes(key, username),
-        decryptBytes(key, password)
-      );
-
-      handler.onDecrypt(results, null);
-    } catch (final UserNotAuthenticatedException ex) {
-      Log.d(LOG_TAG, "Unlock of keystore is needed. Error: " + ex.getMessage(), ex);
-
-      // expected that KEY instance is extracted and we caught exception on decryptBytes operation
       @SuppressWarnings("ConstantConditions") final DecryptionContext context =
-        new DecryptionContext(safeAlias, key, password, username);
+      new DecryptionContext(safeAlias, key, password, username);
 
       handler.askAccessPermissions(context);
     } catch (final Throwable fail) {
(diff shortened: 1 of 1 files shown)


araratispiroglu (approved):

hraschan:
Would also fix #593

alexborton:
Any movement on this issue as we are experiencing it too. Would be nice to have the fix implemented out the box

araratispiroglu:
I've had a similar issue with this library. For me, this was happening when the app was in the foreground and the phone was locked, unlocking the phone was allowing access to the keychain item without secondary biometric verification from the app context. Currently, I'm automatically patching the file after the node modules installation. Adding `.setUserAuthenticationParameters(5, (int) KeyProperties.AUTH_BIOMETRIC_STRONG)` helps me to solve the bug.

ilyassio (approved):

lewie9021:
It worth noting [`.setUserAuthenticationParameters`](https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec.Builder#setUserAuthenticationParameters(int,%20int)) was added in API 30 (Android 11). This will mean increasing the minimum supported version of the library. A more progressive approach might be to assert the version as we build the spec:

```java
final int validityDuration = 5;
final KeyGenParameterSpec.Builder keyGenParameterSpecBuilder = new KeyGenParameterSpec.Builder(alias, purposes)
  .setBlockModes(BLOCK_MODE_ECB)
  .setEncryptionPaddings(PADDING_PKCS1)
  .setRandomizedEncryptionRequired(true)
  .setUserAuthenticationRequired(true)
  .setKeySize(keySize);

if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.R) {
  keyGenParameterSpecBuilder.setUserAuthenticationParameters(validityDuration, KeyProperties.AUTH_BIOMETRIC_STRONG);
} else {
  keyGenParameterSpecBuilder.setUserAuthenticationValidityDurationSeconds(validityDuration);
}

return keyGenParameterSpecBuilder;
```

alemarra89:
Hi all, can I ask you why default validityDuration is 5?
I notice that KeyProperties.AUTH_BIOMETRIC_STRONG value is 2, as documented here:
https://developer.android.com/reference/android/security/keystore/KeyProperties#AUTH_BIOMETRIC_STRONG

lewie9021:
I believe historically this has been lower than 5 seconds but there's been issues with older devices where they simply aren't fast enough to decrypt the data so methods like `getGenericPassword` fail every time

lewie9021:
The value you're referring to is simply the enum's underlying constant value. It's not that it sets it to 2 seconds from what I understand. That is handled when building the `KeyGenParameterSpec` via methods such as `setUserAuthenticationParameters` and `setUserAuthenticationValidityDurationSeconds`

AntoineThibi:
I have used your solution @lewie9021 to keep compatibility with Android SDK < 30. It works great !
For SDK < 30, I just disable the option to access biometry for 5 seconds after coming back to the app. It's not great but I guess there are no real way to insure that the biometry is used to unlock the keystore on those version...

P0labrD:
🆙 @oblador is there a reason not to approve and merge these changes? 🙂

PR #606, opened 2023-08-13
[Fix][Android] UI Blocking on Lower-End Devices

EdgarJMesquita:
### Asynchronous Refactor of `setGenericPassword` to Address UI Blocking on Lower-End Devices

**Issue:**
On certain lower-end devices, a critical issue was identified where the current implementation of the `setGenericPassword` function caused significant UI blocking. This resulted in unresponsive user interfaces lasting up to 22 seconds, greatly impacting user experience and frustrating our users.

**Solution:**
To mitigate this problem and ensure a seamless user experience across all devices, a comprehensive refactor of the `setGenericPassword` function has been undertaken. The primary goal of this refactor was to introduce asynchronous execution, enabling heavy processing tasks to be offloaded to separate threads.

**Impact:**
By implementing this change, we have effectively eliminated the UI blocking issue on lower-end devices. This enhancement not only resolves the reported user problems but also elevates the overall performance and usability of our application on a wider range of hardware.

**Benefits:**
1. **Improved Responsiveness:** Asynchronous execution prevents UI blocking, resulting in a smoother and more responsive user interface, even on resource-constrained devices.
2. **Enhanced User Experience:** Users will no longer face extended periods of unresponsiveness, leading to increased satisfaction and retention.
3. **Optimized Performance:** The updated function execution maintains the security and reliability aspects while delivering better overall application performance.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.java
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.java
@@ -37,6 +37,8 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import javax.crypto.Cipher;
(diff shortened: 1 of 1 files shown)


DorianMazur:
Fixed in #658

PR #610, opened 2023-09-04
Added fallback mechanism for android

gokulkulkarni1987:
### `authenticationType` was not being respected on Android

- After going through the following [link](https://developer.android.com/reference/androidx/biometric/BiometricPrompt.PromptInfo.Builder#setAllowedAuthenticators(int)) found out that there is a scope for adding support for fallback on Android API 30 & above.
- Updated the Example App to support the same.

### DIFF
--- a/KeychainExample/App.js
+++ b/KeychainExample/App.js
@@ -13,7 +13,7 @@ import SegmentedControlTab from 'react-native-segmented-control-tab';
 import * as Keychain from 'react-native-keychain';
 
 const ACCESS_CONTROL_OPTIONS = ['None', 'Passcode', 'Password'];
-const ACCESS_CONTROL_OPTIONS_ANDROID = ['None'];
+const ACCESS_CONTROL_OPTIONS_ANDROID = ['None', 'Fallback'];
 const ACCESS_CONTROL_MAP = [
   null,
   Keychain.ACCESS_CONTROL.DEVICE_PASSCODE,
@@ -22,6 +22,7 @@ const ACCESS_CONTROL_MAP = [
 ];
 const ACCESS_CONTROL_MAP_ANDROID = [
   null,
+  Keychain.ACCESS_CONTROL.BIOMETRY_ANY_OR_DEVICE_PASSCODE,
   Keychain.ACCESS_CONTROL.BIOMETRY_CURRENT_SET,
 ];
 const SECURITY_LEVEL_OPTIONS = ['Any', 'Software', 'Hardware'];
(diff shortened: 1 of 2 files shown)


gokulkulkarni1987:
Guys this is my first PR,

1. it is showing as merging is blocked
2. Some checks haven't completed yet.

I am not sure if I have to do anything more from my side.

If I run `lint` and `flow` locally I do not get any errors,

Screenshot 2023-09-05 at 1 24 17 AM

saadi-ninjasCode:
@gokulkulkarni1987 , I tried your solution but I got the following error
`{"Authentication error is => ": {"error": [Error: Negative text must not be set if device credential authentication is allowed.]}}`

I tested on two Android version 11 & 12

saadi-ninjasCode:
> @gokulkulkarni1987 , I tried your solution but I got the following error `{"Authentication error is => ": {"error": [Error: Negative text must not be set if device credential authentication is allowed.]}}`
> 
> I tested on two Android version 11 & 12
> 
> Screenshot 2023-09-22 at 3 20 40 PM

I fixed this issue by adding `authenticationPrompt: { cancel: '' }` . Furthermore, it should be handled internally in **KeychainModule.java** at line # **622** .  
Screenshot 2023-09-22 at 3 48 19 PM

_ERROR_  => When I add the above line, I get the following error on devices ANDROID_VERSION < 11
` {"Authentication error is => ": {"error": [Error: Negative text must be set and non-empty.]}}`

Expected Behavior: 
It should be working on ANDROID_VERSION < 11. I don't have depth knowledge of ANDROID, But I figured out that if make some changes at line # **622** (Code of block is mentioned below). It should be resolved.

`  if (null != promptInfoOptionsMap && promptInfoOptionsMap.hasKey(AuthPromptOptions.CANCEL)) {
      String promptInfoNegativeButton = promptInfoOptionsMap.getString(AuthPromptOptions.CANCEL);
      promptInfoBuilder.setNegativeButtonText(promptInfoNegativeButton);
    }`

gokulkulkarni1987:
@saadi-ninjasCode We get `Negative text must be set and non-empty` when just `BIOMETRIC_STRONG` is enabled.
The fix I have added is for supporting both `BIOMETRIC_STRONG` & `DEVICE_CREDENTIAL`, which is available only after `Android API 30`, for `DEVICE_CREDENTIALS` to work we must not set cancel text (thats internal of Android).
Below is the screenshot from class `BiometricPrompt`

Screenshot 2023-09-22 at 8 28 20 PM

if you see the lines from `KeychainModule`,

Screenshot 2023-09-22 at 8 36 15 PM

Negative button text setting or not is already handled, so on react-native side if we just check the android version and accordingly pass or not would be better idea.

Issue #593, opened 2023-04-04
getGenericPassword does not require authentication if the phone's Keychain was recently unlocked from another biometric authentication event


ShepSims:
The opening screen of my application tries the following immediately upon launch or forgrounding

const result = await Keychain.getGenericPassword({
          service: 'myapp',
          authenticationPrompt: {
            title: 'Biometric Sign In',
            subtitle: 'Confirm biometrics to continue.',
          },
});
        

This works perfectly in *almost* all cases, however, if my app is in the foreground when the user unlocks their phone, it appears that the biometric success is persisted for a few moments, and thus they are not prompted again for their biometrics to unlock the Keychain.  

Wrapping the Keychain statement in a timeout that forces the app to wait 1 second before trying to get the item from storage as below does fix this issue, however, I don't believe this should be necessary.

 setTimeout(async () => {
          const result = await Keychain.getGenericPassword({
            service: 'molo',
            authenticationPrompt: {
              title: 'Biometric Sign In',
              subtitle: 'Confirm biometrics to continue.',
            },
          });
},1000}
          

This appears to be a fairly large security flaw, especially and I was wondering if anyone else had experienced it, or had any ideas as to why this is happening in the first place.

Abhisflyingsoon:
Experiencing same issue on Android!

PR #613, opened 2023-09-26
use setUserAuthenticationParameters for android >= 11

AntoineThibi:
Fix for Android >= 11 to force the use of biometry if required : an attacker could bypass the use of biometry by unlocking the device with the pin and then could access the keychain, even though we would like it to be protected with the biometry.

Fix coming from a comment on this PR : https://github.com/oblador/react-native-keychain/pull/592

It should solve for Android >=11 the issues : 
- https://github.com/oblador/react-native-keychain/issues/593
- https://github.com/oblador/react-native-keychain/issues/518

However, it does not solve the issue for Android < 11 due to the way the keychain is implemented on Android.


PAN740623:
固定

androideveloper (approved):

renchap:
This seems to be causing issues on Android, where start times are up by 8-20 seconds and even sometimes cause ANRs , see https://github.com/oblador/react-native-keychain/issues/630

PR #622, opened 2024-02-20
feat: add visionOS support

thiagobrez:
This PR adds support for [React Native visionOS](https://github.com/callstack/react-native-visionos), the new out-of-tree platform.

Vision OS has the new Optic ID biometric, but it behaves same as on iOS. All APIs used here are supported, so it was just a matter of considering `TARGET_OS_VISION` in the same preprocessor directives that already had `TARGET_OS_IOS`.

If you think this is good, I can update the docs with visionOS support in a next commit :)

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,9 +22,9 @@
     - [`hasInternetCredentials(server)`](#hasinternetcredentialsserver)
     - [`getInternetCredentials(server, [{ authenticationPrompt }])`](#getinternetcredentialsserver--authenticationprompt-)
     - [`resetInternetCredentials(server)`](#resetinternetcredentialsserver)
-    - [`requestSharedWebCredentials()` (iOS only)](#requestsharedwebcredentials-ios-only)
-    - [`setSharedWebCredentials(server, username, password)` (iOS only)](#setsharedwebcredentialsserver-username-password-ios-only)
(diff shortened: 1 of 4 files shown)


androideveloper:
Hi @thiagobrez ! Great, thanks for the contribution! 

Sounds good to me, I've checked with an iOS engineer 👍 Could you add some docs? 🙏

thiagobrez:
@androideveloper Awesome! Just pushed a commit mentioning visionOS in the docs + 1 missing type for `BIOMETRY_TYPE.OPTIC_ID`. Let me know what you think 🙌

androideveloper:
thanks! @oblador could you check this as well before I merge? 🙏

oblador (approved):
LGTM, thanks for your contribution 👍

androideveloper:
@thiagobrez could you please check this one?
https://github.com/oblador/react-native-keychain/issues/624
Seems to be a breaking change after vision os was added

thiagobrez:
@androideveloper Sure, I was following the same approach on that code block but `@available(visionOS)` happens to only be available on Xcode 15, missed that.

Sent a fix here: https://github.com/oblador/react-native-keychain/pull/631

PR #631, opened 2024-03-22
fix: unrecognized platform name visionOS

### LINKED ISSUE #624: Unrecognized platform name visionOS error
The [last update](https://github.com/oblador/react-native-keychain/compare/v8.1.3...v8.2.0) seems to have created breaking changes. When bumping from `8.1.3` to `8.2.0` and building, Xcode is not able to make it work.

Screenshot 2024-03-01 at 16 54 51

I'm getting the following errors:

- `Unrecognized platform name visionOS`
- `Use of undeclared identifier 'LABiometryTypeOpticID'`

**Config**

- Xcode `14.2`
- RN `0.73.5`

thiagobrez:
Fixes #624 

Remove `@available(visionOS 1, *)`, which is only available on Xcode 15 and above. Use obj-c preprocessor directives instead.

This was following the same approach as iOS in the lines below `@available(iOS 11, *)`. iOS needs this because FaceID is only available on iOS 11 and above. VisionOS does not need it because visionOS 1 already has OpticID since launch.

Tested on Xcode 14.3.1 and 15.2.

### DIFF
--- a/RNKeychainManager/RNKeychainManager.m
+++ b/RNKeychainManager/RNKeychainManager.m
@@ -333,11 +333,11 @@ RCT_EXPORT_METHOD(getSupportedBiometryType:(RCTPromiseResolveBlock)resolve
   BOOL canBeProtected = [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:&aerr];
 
   if (!aerr && canBeProtected) {
-    if (@available(visionOS 1, *)) {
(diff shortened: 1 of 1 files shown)


androideveloper (approved):

androideveloper:
fyi @oblador new release is needed for this 🙏

jspizziri:
@oblador is there anyone who can release this project other than you?

PR #639, opened 2024-05-31
NAT-171 Fix message backup - add Blockstore storage for Keychain lib

nowakowski-damian:
Add Blockstore storage for Keychain lib

  Google BlockStore API storage -> https://developers.google.com/identity/blockstore/android

  IMPORANT NOTES:
  1) Block Store data is persisted across the app uninstall/reinstall only when the user enables Backup services (it can be checked at Settings > Google > Backup)
  2) Cloud backup and restore is enabled only when:
       - End to End encryption is enabled
       - Source device runs API 23+
       - Target device runs API 31+ (API 29+ for Pixel phones)
  3) E2E encryption is supported:
       - on devices running Android 9 (API 29) and above
       - when device have a screen lock set with a PIN, pattern, or password


android/src/main/java/com/oblador/keychain/KeychainModule.java
```diff
@@ -130,23 +131,24 @@ public class KeychainModule extends ReactContextBaseJavaModule {
...
+//    addCipherStorageToMap(new CipherStorageFacebookConceal(reactContext));
     addCipherStorageToMap(new CipherStorageKeystoreAesCbc());
 
     // we have a references to newer api that will fail load of app classes in old androids OS
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
-      addCipherStorageToMap(new CipherStorageKeystoreRsaEcb());
-    }
+//    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+//      addCipherStorageToMap(new CipherStorageKeystoreRsaEcb());
+//    }
```
nowakowski-damian, line 151: I commented
` //    addCipherStorageToMap(new CipherStorageFacebookConceal(reactContext));`
and
```
//    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
//      addCipherStorageToMap(new CipherStorageKeystoreRsaEcb());
//    }
```
as it was commented originally in our patch.

Issue #533, opened 2022-04-10
IOS key persists after uninstall??


IKTANIM:
IOS key persists even if uninstall the app?? What  to do regarding this??

Adnan-Bacic:
i have previously used react-native-encrypted-storage and i remember this. they included it in their README file. i would think its the same solution here, though i havent tried it:

https://github.com/emeraldsanto/react-native-encrypted-storage#note-regarding-keychain-persistence

BraveEvidence:
This will help https://www.youtube.com/watch?v=J0OSn7s9YiA&list=PLQhQEGkwKZUrempLnmxjt7ZCZJu1W3p2i&index=15

Issue #135, opened 2018-06-27
Android: App uninstall wipes out credentials


a-koka:
Seems like my credentials are being wiped out on Android upon app uninstall on Android 6, 7 & 8. Am I missing something here? I was under the impression that keychain data is retained even after an uninstall. Seems to work this way on iOS. Any suggestions would be appreciated.

MacKentoch:
- **iOS** won't wipe keychain app associated data on app uninstall.
But by default this is not saved on iCloud but on device itself so if you change your device you won't find the data.

- **Android** won't persist data (this does not work like iOS keychain).
In Android maybe `android:allowBackup="true"` (I did not try) should do the trick. In `AndroidManifest.xml`:

```xml
<application
      android:name=".MainApplication"
      android:label="@string/app_name"
      android:icon="@mipmap/ic_launcher"
      android:allowBackup="true"
```

a-koka:
`allowBackup` is already set to `true` for my app. Looks like I was operating with the misunderstanding of the keychain being persisted even after an app uninstall.

With a combination of `AsyncStorage` and some simple encryption algorithm, one can easily achieve this functionality on both platforms. Would it be possible to shed some light on the use cases where this library comes in handy on an Android device given that the data isn't persisted?

leethree:
This behaviour breaks the feature parity between iOS and Android. People would expect that the library works the same way on both platforms. So I think it should be fixed in this library (i.e. wipe out credentials on both iOS and Android) or at least documented.

Related issue: #87

hellogerard:
I have been looking into this as well. It does not look like we can get persistent "keychain" data  on Android after an app uninstall. The encrypted data itself is stored in SharedPreferences which is by default deleted on app uninstall. In theory, this could get backed up/restored using Android's Auto Backup feature (though I could not get even this to work). However, the keystore entries also get deleted on app uninstall, so even if we had the data, we could not decrypt them without the keys. I think I've got that right.

> With a combination of AsyncStorage and some simple encryption algorithm, one can easily achieve this functionality on both platforms. 

I'm not a security expert, but I'd like to try to answer this for my own edification because I've been asking myself similar questions lately.

[...]

PR #642, opened 2024-06-24
feat: hasGenericPassword

DorianMazur:
## Description

This new function `hasGenericPasswordForOptions` for both iOS and Android platforms. This function checks if a generic password exists for a given service and returns a Promise<boolean>.

## Motivation and Context

In some applications, it's crucial to verify the existence of data in the Keychain (iOS) or Shared Preferences (Android) before attempting to access it (for example when you store encryption key to decrypt the data) This function provides a convenient way to perform this check, similar to the existing `hasInternetCredentials` function. 

Also it should help to solve the issue #135 & #533 
We will have a way to determine if the data is in Keychain.

## How Has This Been Tested?

Manually tested in a real React Native project on iOS & Android.


DorianMazur:
@oblador @vonovak Can you check this PR? Is the project still maintained?

Issue #567, opened 2022-08-31
`E_CRYPTO_FAILED` on Android 13 and Pixel phones. AES crashing out while decrypting.


kaushil111:
Stack trace:
```
{
  "nativeStackAndroid": [
    {
      "lineNumber": 150,
      "file": "CipherStorageKeystoreAesCbc.java",
      "methodName": "decrypt",
      "class": "com.oblador.keychain.cipherStorage.CipherStorageKeystoreAesCbc"
    },
    {
      "lineNumber": 165,
      "file": "CipherStorageKeystoreAesCbc.java",
      "methodName": "decrypt",
      "class": "com.oblador.keychain.cipherStorage.CipherStorageKeystoreAesCbc"
    },
    {
      "lineNumber": 679,
      "file": "KeychainModule.java",
      "methodName": "decryptToResult",
      "class": "com.oblador.keychain.KeychainModule"
    },
    {
      "lineNumber": 646,
      "file": "KeychainModule.java",
      "methodName": "decryptCredentials",
      "class": "com.oblador.keychain.KeychainModule"
    },
    {
      "lineNumber": 306,
      "file": "KeychainModule.java",
      "methodName": "getGenericPassword",
      "class": "com.oblador.keychain.KeychainModule"
    },
    {
      "lineNumber": 367,
      "file": "KeychainModule.java",
      "methodName": "getGenericPasswordForOptions",
      "class": "com.oblador.keychain.KeychainModule"
    },
    {
      "lineNumber": -2,
      "file": "Method.java",
      "methodName": "invoke",
      "class": "java.lang.reflect.Method"
    },
    {
      "lineNumber": 372,
      "file": "JavaMethodWrapper.java",
      "methodName": "invoke",
      "class": "com.facebook.react.bridge.JavaMethodWrapper"
    },
    {
      "lineNumber": 188,
      "file": "JavaModuleWrapper.java",
      "methodName": "invoke",
      "class": "com.facebook.react.bridge.JavaModuleWrapper"
    },
    {
      "lineNumber": -2,
      "file": "NativeRunnable.java",
      "methodName": "run",
      "class": "com.facebook.react.bridge.queue.NativeRunnable"
    },
    {
      "lineNumber": 942,
      "file": "Handler.java",
      "methodName": "handleCallback",
      "class": "android.os.Handler"
    },
    {
      "lineNumber": 99,
      "file": "Handler.java",
      "methodName": "dispatchMessage",
      "class": "android.os.Handler"
[...]

Issue #372, opened 2020-07-16
Retrieving password on android fails with Caused by: android.security.KeyStoreException: Unsupported key size


mihaelaLo:
I am storing the accessToken and refreshToken information. Retrieving them fails on android with the below problem. All works ok on iOS. Also downgrading to version ^4.0.5 solves the problem on Android also.

```
2020-07-16 18:21:09.746 594-594/? E//vendor/bin/hw/android.hardware.keymaster@4.0-service.citadel: GenerateKey : device response error code: UNSUPPORTED_KEY_SIZE
2020-07-16 18:21:09.748 954-2883/? D/DropBoxManager: About to call service->add()
2020-07-16 18:21:09.749 2785-2864/com.app.myapp E/KeyStore: generateKeyInternal failed on response -6
2020-07-16 18:21:09.749 954-954/? I/keystore: del USRPKEY_warmingUp 10519
2020-07-16 18:21:09.749 954-954/? I/keystore: del USRSKEY_warmingUp 10519
2020-07-16 18:21:09.749 954-954/? I/keystore: del USRCERT_warmingUp 10519
2020-07-16 18:21:09.750 954-954/? I/keystore: del CACERT_warmingUp 10519
2020-07-16 18:21:09.751 2785-2864/com.app.myapp W/CipherStorageBase: StrongBox security storage is not available.
    java.security.ProviderException: Failed to generate key pair
        at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeystoreKeyPair(AndroidKeyStoreKeyPairGeneratorSpi.java:514)
        at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPair(AndroidKeyStoreKeyPairGeneratorSpi.java:470)
        at java.security.KeyPairGenerator$Delegate.generateKeyPair(KeyPairGenerator.java:727)
        at com.oblador.keychain.cipherStorage.CipherStorageKeystoreRsaEcb.generateKey(CipherStorageKeystoreRsaEcb.java:256)
        at com.oblador.keychain.cipherStorage.CipherStorageBase.tryGenerateStrongBoxSecurityKey(CipherStorageBase.java:444)
        at com.oblador.keychain.cipherStorage.CipherStorageBase.generateKeyAndStoreUnderAlias(CipherStorageBase.java:391)
        at com.oblador.keychain.KeychainModule.internalWarmingBestCipher(KeychainModule.java:173)
[...]

PR #647, opened 2024-08-07
fix!: remove padding from decryptedBytes if left in by the cipher

andrejborstnik:
On (some?) Pixel devices (tested Pixel 6 and Pixel 7 on Android 14, `AP2A.240705.004` build) for certain string lengths (e.g. `256`) the encryption/decryption using `CipherStorageKeystoreAesCbc` doesn't erase the `PKCS7` padding from the returned string.

To work around this issue I added a method that will detect & delete the padding if left in.

This technically a breaking change for any stored strings that end with a valid `PKCS7` padding of characters up to with an ascii int value of `IV.IV_LENGTH`. Not sure how likely this use-case would be.

Opened this workaround as a discussion point, happy to discuss alternatives.

Also reported to Android https://issuetracker.google.com/issues/357896946, so they can look at the issue on their side

### DIFF
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
@@ -228,12 +228,32 @@ public class CipherStorageKeystoreAesCbc extends CipherStorageBase {
     return generator.generateKey();
   }
 
+  private static byte[] maybeRemovePKCS7Padding(byte[] paddedBytes, int maxPaddingLength) {
+    int paddingLength = paddedBytes[paddedBytes.length - 1];
+
+    // Validate the padding
+    if (paddingLength < 1 || paddingLength > paddedBytes.length || paddingLength > maxPaddingLength) {
+      return paddedBytes;
+    }
+    for (int i = paddedBytes.length - paddingLength; i < paddedBytes.length; i++) {
+      if (paddedBytes[i] != paddingLength) {
+        return paddedBytes;
+      }
+    }
+
+    // Remove the padding
+    byte[] unpaddedBytes = new byte[paddedBytes.length - paddingLength];
+    System.arraycopy(paddedBytes, 0, unpaddedBytes, 0, unpaddedBytes.length);
+
+    return unpaddedBytes;
+  }
+
   /** Decrypt provided bytes to a string. */
   @NonNull
   @Override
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageKeystoreAesCbc.java
```diff
@@ -228,23 +228,49 @@ protected Key generateKey(@NonNull final KeyGenParameterSpec spec) throws Genera
...
-      // decrypt the bytes using cipher.doFinal(). Using a CipherInputStream for decryption has historically led to issues
+      // decrypt the bytes using cipher.doFinal(). Using a CipherInputStream for
+      // decryption has historically led to issues
       // on the Pixel family of devices.
       // see https://github.com/oblador/react-native-keychain/issues/383
-      byte[] decryptedBytes = cipher.doFinal(bytes, IV.IV_LENGTH, bytes.length - IV.IV_LENGTH);
+      byte[] _decryptedBytes = cipher.doFinal(bytes, IV.IV_LENGTH, bytes.length - IV.IV_LENGTH);
+
+      // removing padding is required to work around a decryption padding issue with
+      // some Pixel devices e.g. for strings of length 256
```
andrejborstnik, line 271: Based on my testing up to 2000 bytes the problematic strings lengths are
```
528 -> 543
800 -> 815
1280 -> 1295
1552 -> 1567
1824 -> 1839

```

jahead:
would this be the reason we are getting 
* https://github.com/oblador/react-native-keychain/issues/372
* https://github.com/oblador/react-native-keychain/issues/567
* and a number of others?

andrejborstnik:
> would this be the reason we are getting
> 
> * [Retrieving password on android fails with Caused by: android.security.KeyStoreException: Unsupported key size #372](https://github.com/oblador/react-native-keychain/issues/372)
> * [`E_CRYPTO_FAILED` on Android 13 and Pixel phones. AES crashing out while decrypting. #567](https://github.com/oblador/react-native-keychain/issues/567)
> * and a number of others?

I have only seen the issue in the OP on Pixel devices and `E_CRYPTO_FAILED` happens on other devices as well. So I would be sceptical this is the only reason. But I would not exclude this being the reason for at least some of the decryption issues, e.g. [this padding one](https://github.com/oblador/react-native-keychain/issues/567#issuecomment-1276762307) seems related on the surface

DorianMazur:
@andrejborstnik @jahead Consider using AES GCM #678

PR #651, opened 2024-08-26
feat: refactor and add e2e tests

DorianMazur:
## Motivation

This PR introduces end-to-end (e2e) tests to ensure the library functions as expected. The existing Java unit tests were outdated (last updated 3 years ago), only covered Android (with no iOS support), and were difficult to maintain. Due to these issues, I decided to remove the unit tests, as they were no longer helpful and were complicating the development process.

image

## Description
This PR includes the following changes:

1. Added e2e tests using Detox: I chose Detox over Maestro because Detox provides better support for biometrics and adb commands, even though both tools have a similar setup complexity (in my opinion)
2. Refactored the codebase to TypeScript: I transitioned the code from Flux to TypeScript, as it is now a widely accepted standard.
3. Reorganized code structure: I moved the Android-specific code to an android directory and the iOS-specific code to an ios directory for better organization.
4. Updated the codebase: The code has been updated to support the latest versions of React Native.
5. Integrated react-native-bob-builder: This tool is now used to build the library.
6. Improved the example app: I refactored the example app to use react-native-test-app, making it easier to update to newer versions of React Native. This also sets the stage for potentially adding e2e tests for different React Native versions.


.github/workflows/lint_and_types.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+      - name: Enable corepack
+        run: corepack enable
+      - name: Install dependencies
+        run: yarn install --immutable
+      - name: Check flow types
```
oblador, line 42:
```suggestion
      - name: Check types
```

KeychainExample/App.tsx
```diff
@@ -44,33 +45,36 @@ export default class KeychainExample extends Component {
...
+    biometryType: undefined,
+    accessControl: undefined as undefined | Keychain.ACCESS_CONTROL,
+    securityLevel: undefined as undefined | Keychain.SECURITY_LEVEL,
+    storage: undefined as undefined | Keychain.STORAGE_TYPE,
+    selectedStorageIndex: 0,
+    selectedSecurityIndex: 0,
+    selectedAccessControlIndex: 0,
   };
 
   componentDidMount() {
```
oblador, line 57: Could consider refactoring this one to a functional component at some point.

KeychainExample/App.tsx
```diff
@@ -165,16 +169,16 @@ export default class KeychainExample extends Component {
         style={styles.container}
       >
         <View style={styles.content}>
-          <Text style={styles.title}>Keychain Example</Text>
+          <Text style={styles.title} onPress={() => Keyboard.dismiss()}>
```
oblador, line 172: Seems odd, maybe makes more sense to wrap the whole thing in a scrollview to handle taps to dismiss?

src/index.ts
```diff
@@ -180,7 +161,9 @@ export function setGenericPassword(
   password: string,
   serviceOrOptions?: string | Options
 ): Promise<false | Result> {
+  console.log('setGenericPassword', serviceOrOptions);
   const options = normalizeOptions(serviceOrOptions);
+  console.log('normalized options', options);
```
oblador, line 166:
```suggestion
  const options = normalizeOptions(serviceOrOptions);
```

android/src/test/java/com/oblador/keychain/KeychainModuleTests.java
oblador, line 1: I feel like there are quite a few cases in this test suite not covered by the e2e from what I can tell. Can we cover them too (esp around which crypto engine picked and migration paths)?

KeychainExample/App.tsx
DorianMazur: Yes, that is the plan eventually :D

KeychainExample/App.tsx
DorianMazur: This is a quick hack to hide the keyboard for end-to-end tests. I wouldn't worry about it, but I can't find a better way to do it, if this is important

android/src/test/java/com/oblador/keychain/KeychainModuleTests.java
DorianMazur: Are you talking about RSA, AES and migration from AES to RSA?

KeychainExample/App.tsx
oblador: Ah, makes sense. In a production app I would have cared, but not in a tester app 👍

android/src/test/java/com/oblador/keychain/KeychainModuleTests.java
oblador: Yeah

DorianMazur: Done! 👍 Once the pipeline is finished, I'll merge it.

PR #655, opened 2024-09-21
ci: add code coverage

Killpit:
-Added code coverage tooling for to check code coverage and quality
-Planning to add Snyk for security scans, and they both might need tokens and membership, I can do the latter for the project.


.github/workflows/e2e_tests.yaml
```diff
@@ -123,6 +123,12 @@ jobs:
             yarn test:android:run
         env:
           API_LEVEL: ${{ matrix.api-level }}
+
+      - name: Test coverage
```
DorianMazur, line 127: If we want to add this, then we need this for Android and IOS. Not only for Android.

DorianMazur:
Can you run the workflow and show that it's working? I checked your repo, and the workflow seems to be failing there: -> https://github.com/Killpit/react-native-keychain/actions/runs/10971100884

Killpit:
> Can you run the workflow and show that it's working? I checked your repo, and the workflow seems to be failing there: -> https://github.com/Killpit/react-native-keychain/actions/runs/10971100884

Will do, they rarely allow running CI/CD, which I'll run it and fix the problems accordingly

.github/workflows/e2e_tests.yaml
Killpit: Will do both, thanks for the feedback

PR #682, opened 2024-11-12
Default value for authenticationPrompt

dlindstrm:
If no value is passed for authenticationPrompt when settings internet credentials or generic password it throws an error: `Title must be set and non-empty`. This is a breaking change since v8.


DorianMazur:
I have already created a PR to address this issue: https://github.com/oblador/react-native-keychain/pull/683/files.
Nevertheless, thank you for your PR, @dlindstrm!

PR #684, opened 2024-11-12
setInternetCredentials default value for authenticationPrompt

dlindstrm:
If no options is passed to setInternetCredentials it will throw an error.


DorianMazur:
Ah, I missed that. Thanks!

PR #685, opened 2024-11-13
fix: IllegalBlockSizeException on Android while decrypting

DorianMazur:
This fixes the javax.crypto.IllegalBlockSizeException that sometimes occurs during decryption.

```
javax.crypto.IllegalBlockSizeException
java.io.IOException: javax.crypto.IllegalBlockSizeException
at javax.crypto.CipherInputStream.getMoreData(CipherInputStream.java:133)
at javax.crypto.CipherInputStream.read(CipherInputStream.java:249)
at javax.crypto.CipherInputStream.read(CipherInputStream.java:225)
```

### DIFF
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.kt
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherStorageBase.kt
@@ -1,10 +1,12 @@
 package com.oblador.keychain.cipherStorage
 
+import android.annotation.SuppressLint
 import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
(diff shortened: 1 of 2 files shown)


SYoder1:
@DorianMazur I am seeing this same issue on 9.1.0 using biometrics RSA. Do you know if this should fix that too? It looks like it is part of the base storage class that RSA extends from.

DorianMazur:
@SYoder1 please upgrade to 10.x

SYoder1:
I can't upgrade to v10 yet, because of the #730 issue

DorianMazur:
@SYoder1 #730 is not 10.x only...

Issue #704, opened 2024-12-16
`getGenericPassword` failing after updating package from 8.2.0


Bowlerr:
**Title:** Incompatibility Between Versions 8.2.0 and 9.2.2 with `getGenericPassword`

**Description:**  
When upgrading from `react-native-keychain` version `8.2.0` to `9.2.2`/`9.2.1`, I encountered an error when calling `getGenericPassword`. The password was set using version `8.2.0`, but when attempting to retrieve it in version `9.2.2`, the following error occurs:

```
Wrapped error: The message could not be decrypted successfully. It has either been tampered with or the wrong resource is being decrypted.
```

**Steps to Reproduce:**  
1. Set the password using version `8.2.0`:
    ```javascript
    setGenericPassword('pin', encrypted, {
      service: 'passcode',
      storage: STORAGE_TYPE.FB,
    });
    ```
2. Upgrade to version `9.2.2` or `9.2.1`:
3. Attempt to retrieve the password using:
    ```javascript
    await getGenericPassword({ service: 'passcode' });
    ```

**Observed Behavior:**  
The `getGenericPassword` call fails with the error:
```
Wrapped error: The message could not be decrypted successfully. It has either been tampered with or the wrong resource is being decrypted.
```

**Expected Behavior:**  
The `getGenericPassword` call should successfully retrieve the password set in version `8.2.0`.

**Environment:**  
- **Devices Tested:** 
  - OnePlus 12 (Android 14)  
  - Pixel 7a (Android 15)  
- **React Native Keychain Version:** 
  - Set: `8.2.0`  
  - Retrieve: `9.2.2`  
- **React Native Version:** `0.74.5` 

**Additional Notes:**  
It seems like the encryption format or decryption logic may have changed between the two versions, resulting in backward compatibility issues. If this is expected, could you provide guidance on migrating securely without data loss?

DorianMazur:
Hi! I’ll take a look at this soon. Just one quick question: is there a specific reason you’re still using FacebookConceal? It’s outdated and will be removed soon. Are other ciphers affected as well?

Bowlerr:
> Hi! I’ll take a look at this soon. Just one quick question: is there a specific reason you’re still using FacebookConceal? It’s outdated and will be removed soon. Are other ciphers affected as well?

@DorianMazur I found this issue whilst upgrading from `FacebookConceal` to `AES_GCM_NO_AUTH` or `AES_GCM` depending on if the user wants to use biometrics.

[...]

PR #707, opened 2025-01-16
Fix: Resolve Incompatibility Between react-native-keychain Versions 8.2.0 and 9.0.0 for getGenericPassword

Bowlerr:
Resolves [[#704](https://github.com/oblador/react-native-keychain/issues/704)]

### **Description:**  

This PR addresses an incompatibility issue in `react-native-keychain` between versions 8.2.0 and 9+. The issue arises when passwords set using version 8.2.0 cannot be retrieved using version 9+ due to differences in `Entity` prefix construction during encryption and decryption.

#### Issue Summary:  
- When attempting to retrieve passwords set with version 8.2.0 using `getGenericPassword` in version 9+, the following error is encountered:
  ```
  Wrapped error: The message could not be decrypted successfully. It has either been tampered with or the wrong resource is being decrypted.
  ```

#### Root Cause:  
- A discrepancy in the `Entity` prefix construction caused a mismatch between how encrypted data was stored in version 8.2.0 and how it is retrieved in version 9+

---

### **Changes Made:**  
1. Removed an unintentional space in FC's `createPasswordEntity` and `createUsernameEntity` methods in the Kotlin implementation to ensure consistent prefix generation with the Java implementation.
2. Ensured the prefix format matches the version 8.2.0 logic:
   - Java: `RN_KEYCHAIN:<alias>pass`
   - Kotlin: `RN_KEYCHAIN:<alias>pass` (now consistent).

---

### **Impact:**  
- Resolves the compatibility issue between versions 8.2.0 and future versions.
- Ensures that passwords encrypted in older versions can still be decrypted successfully in newer versions.

### **Environment Details:**
- **Devices Tested:**
  - OnePlus 12 (Android 14)
  - Pixel 7a (Android 15)
- **React Native Keychain Version:**
  - Password Set: 8.2.0
  - Password Retrieve: 9.2.2
- **React Native Version:** 0.74.5

---

### **Checklist:**  
- [x] Verified encryption and decryption compatibility across versions.
- [x] Reviewed prefix generation logic in both Java and Kotlin.
- [x] Tested on relevant devices and Android versions.


Bowlerr:
Checks failed from [here](https://github.com/oblador/react-native-keychain/commit/e9f6ac486add14c3477c762cee6beb1fb5ec921a)

DorianMazur:
@Bowlerr Great work 💯 These things are the easiest to miss.
I will test this PR and investigate why the Gradle build is failing. I suspect this change might affect users who have upgraded react-native-keychain to version 9.x.x and still use FacebookConceal. However, since FacebookConceal is deprecated and we are planning to remove it, I believe this should not be a major issue. Merging this PR will help ease the transition for users coming from older versions of react-native-keychain.

Bowlerr:
@DorianMazur that was a concern of mine too. 
We could check both prefixes if one fails however I do not know if this will cause any issues with performance

Bowlerr:
@DorianMazur Managed to fix the gradle by updating kotlin and it built fine but I need to test it further 
https://github.com/oblador/react-native-keychain/pull/709

Bowlerr:
@DorianMazur All tests pass locally after - https://github.com/oblador/react-native-keychain/pull/709

PR #712, opened 2025-01-22
feat: getAllGenericPasswordServices allow skipping items requiring UI auth on iOS

jeanregisser:
We need to list all keychain items in our iOS app, however items that are protected with biometry (Touch ID, Face ID, etc) would trigger the auth prompt upon listing.

This adds an option to silently skip such items.
See https://developer.apple.com/documentation/security/ksecuseauthenticationuiskip?language=objc

This change is fully backward compatible.

Note: we were using a [custom patch](https://github.com/valora-inc/wallet/blob/299072a53833e8859137e56ecccf5220630891a5/patches/react-native-keychain%2B8.2.0.patch) for this, but we'd love to have this option in the official release. Hence this PR ;)

Let me know if you'd like anything adjusted.


DorianMazur:
Hey @jeanregisser, thank you! I've been a bit busy lately, but I'll try to review it soon.

PR #714, opened 2025-02-10
fix: update Android `react-native` resolution approach in monorepos

bakoushin:
### Description

Update the React Native package resolution in Android for monorepos, aligning with the recent approach used in [`react-native-reanimated`](https://github.com/software-mansion/react-native-reanimated/blob/b703185e9f3ea3dfd5247477177820795c6f23ec/packages/react-native-reanimated/android/build.gradle#L73-L88).

### Rationale

This update introduces a nesting-agnostic resolution approach, making it more adaptable to various monorepo structures. Different projects organize their repositories with varying levels of nesting. For example, in the [`@divvy/mobile`](https://github.com/divvixyz/divvi-mobile) framework, apps are structured within the `/apps` directory rather than at the root. This change ensures better compatibility across different monorepos.

### DIFF
--- a/android/react-native-helpers.gradle
+++ b/android/react-native-helpers.gradle
@@ -4,23 +4,17 @@ def safeAppExtGet(prop, fallback) {
 }
 
 // Let's detect react-native's directory, it will be used to determine RN's version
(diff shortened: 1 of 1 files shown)


DorianMazur:
@bakoushin Thanks!

Issue #730, opened 2025-03-07
com.oblador.keychain.exceptions.CryptoFailedException: Wrapped error: null


phamngocthinh:
I encountered an issue when calling the getGenericPassword method from react-native-keychain.

My usage is:
`await Keychain.setGenericPassword(username, password, { service: USER_KEY123456 });`
`await Keychain.getGenericPassword({ service: USER_KEY123456 });`

Some users randomly experience the following error:
`com.oblador.keychain.exceptions.CryptoFailedException: Wrapped error: null
`

This issue seems device-specific and occurs randomly, making it difficult to reproduce consistently.
Does anyone have experience with this problem or suggestions on how to resolve it?

<img width="660" alt="Image" src="https://github.com/user-attachments/assets/c01536fb-f94a-4694-9e34-2ce0841a7c56" />

DorianMazur:
What version of react-native-keychain are you using?

phamngocthinh:
I’m using version 9.2.2 of react-native-keychain.

DorianMazur:
I've identified the root cause: When using a custom service:
```tsx
Keychain.setGenericPassword(username, password, { service: 'customService' })
```
sometimes the AES_GCM cipher storage is attempting to use the same keystore alias in both ciphers AES_GCM and AES_GCM_NO_AUTH. This alias collision is triggering the error you're experiencing. I am working on a fix.

phamngocthinh:
@DorianMazur  Thank you so much for feedback on this issue,
Could you suggest any temporary workaround that we can apply for affected customers? Also, do you have an estimated timeline for when the fix will be officially released?

Bowlerr:
If I understood the issue correctly, @DorianMazur does this solve the issue? https://github.com/oblador/react-native-keychain/pull/736

SYoder1:
I am seeing the same, but the errors are very heavily weighted to the Motorola Razr 

![Image](https://github.com/user-attachments/assets/40618e7a-be44-454b-aedf-940db915c7df)

DorianMazur:
@Bowlerr @SYoder1 
The prefix might help with some of the issues, but after thinking about it, I don’t believe it’s the root cause of the error. This is a general error, so the first step should be to add better error handling. I’ll look into it further, I think it is an edge case and quite hard to reproduce. Prefix was my first guess as it's something I want to change for quite some time.

Bowlerr:
```
14437:14523 W/c: null
[...]

PR #738, opened 2025-03-19
refactor: Improve Cipher Management with Thread-Safe Caching

Bowlerr:
# Improve Cipher Management with Thread-Safe Caching and Potential Fix for AEADBadTagException

## Summary

This PR improves the management of `Cipher` instances by implementing a new `CipherCache` utility with ThreadLocal-based caching. These changes enhance thread safety, and might help mitigate potential causes of [AEADBadTagException](https://github.com/oblador/react-native-keychain/issues/730). By eliminating shared mutable state, ensuring proper reinitialisation, and cleaning up cached resources, the PR addresses issues such as state leakage that can lead to authentication tag mismatches.

## Changes

- **New CipherCache Utility:**
  - Implemented a new `CipherCache` singleton with ThreadLocal storage to manage `Cipher` instances.
  - Ciphers are cached per transformation string to support different configurations.
- **Thread-Safe Caching:**
  - Replaced the old single-instance caching mechanism in `CipherStorageBase` with ThreadLocal-based caching.
  - Each thread now has its own isolated `Cipher` instance, eliminating synchronisation bottlenecks.
- **Cleanup and Memory Leak Prevention:**
  - Added proper cleanup of cached `Cipher` instances on module invalidation (e.g., in `KeychainModule.invalidate()`).
  - Fixed potential memory leaks by cleaning up ThreadLocal references.
- **Potential AEADBadTagException Mitigation:**
  - Improved cipher initialisation performance and state management ensure that each `Cipher` is reinitialised with the correct parameters (key, IV, etc.) before use.
  - Isolated cipher instances per thread help mitigate errors from concurrent use that can lead to AEADBadTagException.

## Technical Details

- **CipherCache Implementation:**
  - Uses `ThreadLocal<MutableMap<String, Cipher>>` keyed by transformation string.
  - Each thread has its own cache, ensuring that `Cipher` instances are not shared across threads.
- **Proper Reinitialisation:**
  - Cached `Cipher` instances must be reinitialised (via `cipher.init(...)`) before each operation to clear any previous state.
- **Resource Cleanup:**
  - The new cleanup method in `KeychainModule.invalidate()` clears the ThreadLocal cache to prevent memory leaks, especially in long-lived thread environments.
- **Removal of Old Caching Mechanism:**
  - The previous single-instance caching in `CipherStorageBase` has been removed to avoid shared mutable state.

### DIFF
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherCache.kt
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherCache.kt
@@ -1,0 +1,37 @@
+package com.oblador.keychain.cipherStorage
+
+import android.util.Log
+import java.security.NoSuchAlgorithmException
+import javax.crypto.Cipher
+import javax.crypto.NoSuchPaddingException
+
+object CipherCache {
+    private val LOG_TAG = CipherCache::class.java.simpleName
+
+    private val cipherCache = ThreadLocal<MutableMap<String, Cipher>>()
+
+    @Throws(NoSuchAlgorithmException::class, NoSuchPaddingException::class)
+    fun getCipher(transformation: String): Cipher {
+        var ciphers = cipherCache.get()
+        if (ciphers == null) {
+            ciphers = HashMap()
+            cipherCache.set(ciphers)
+        }
+
+        var cipher = ciphers[transformation]
+        if (cipher == null) {
+            cipher = Cipher.getInstance(transformation)
+            ciphers[transformation] = cipher
+        }
+
+        return cipher
+    }
+
+    fun clearCache() {
+        try {
+            cipherCache.remove()
+        } catch (e: Exception) {
+            Log.w(LOG_TAG, "Failed to clear cipher cache: ${e.message}")
+        }
+    }
+}
--- a/android/src/main/java/com/oblador/keychain/cipherStorage/CipherCache.kt
+++ b/android/src/main/java/com/oblador/keychain/cipherStorage/CipherCache.kt
@@ -24,7 +24,7 @@ object CipherCache {
             ciphers[transformation] = cipher
         }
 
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/oblador/keychain/cipherStorage/CipherCache.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    @Throws(NoSuchAlgorithmException::class, NoSuchPaddingException::class)
+    fun getCipher(transformation: String): Cipher {
+        var ciphers = cipherCache.get()
+        if (ciphers == null) {
+            ciphers = HashMap()
+            cipherCache.set(ciphers)
+        }
+
+        var cipher = ciphers[transformation]
```
Bowlerr, line 21: Does transformation include if it requires Auth or not or should I pass the prefix here too after the merge if the prefixes are good to go? @DorianMazur

DorianMazur: No, we shouldn't put things like that into transformation string
https://developer.android.com/reference/javax/crypto/Cipher#getInstance(java.lang.String)

Bowlerr: Sorry for the confusion, I was meaning for the key for the cipher. Does non Auth and Auth have the same transformation string? They may work better on its own cipher thread

DorianMazur: Yes, they have the same transformation string

Bowlerr:
Added an optional prefix so we can separate the instances if needed to avoid shared mutable state. can pass the prefix if that other PR is all good to go

PR #741, opened 2025-03-22
test: improve e2e tests

DorianMazur:


Bowlerr:
### **🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉**

DorianMazur:
@Bowlerr It's not a success yet! :D What I'm trying to do is take a avd snapshot and save it in the repo. This should help fix flaky biometric enrollment, make tests run faster, and let us check if data stored with the old react-native-keychain is still accessible in newer versions.

DorianMazur:
But it's not working so far

Bowlerr:
got excited by seeing the passing tests in the PR tab, didn't look any further into it haha. Sounds fantastic great work!

Bowlerr:
Have a theory that the keyboard may cover the text as it happens usually for the passcode input if somehow left open but was a shower thought and haven't confirmed it

PR #743, opened 2025-03-23
feat: isPasscodeAuthAvailable

DorianMazur:
This pull request includes several changes to improve the handling of passcode and biometric authentication in the `KeychainModule` for Android and iOS, as well as some code cleanup and refactoring.


jacksondoherty:
@DorianMazur Curious the context around this PR – is there a class of Android devices that doesn't support device passcode? Or does this have more to do with if a passcode has been set on the device?

DorianMazur:
@jacksondoherty There really isn’t a modern Android device that cannot present a lock-screen PIN / pattern / password. The check is almost entirely about whether the user already has a credential enrolled and whether the system will let you invoke it right now

PR #754, opened 2025-05-16
Increase min sdk requirement for strongbox

androideveloper:
### Motivation
Increase the sdk level requirement for strongbox usage from android 9 to android 12 in order to improve performance of low-end devices with old android versions. 

Image

### Security impact
Attestation level drops from STRONGBOX to TRUSTED_ENVIRONMENT on devices where a TEE exists. Keys are still hardware-backed inside TrustZone; risk increase is limited to sophisticated physical side-channel attacks.

From Google docs https://developer.android.com/privacy-and-security/keystore
Screenshot 2025-05-16 at 14 11 09

### DIFF
--- a/android/src/main/java/com/oblador/keychain/DeviceAvailability.kt
+++ b/android/src/main/java/com/oblador/keychain/DeviceAvailability.kt
@@ -15,7 +15,7 @@ import androidx.biometric.BiometricManager
 object DeviceAvailability {
 
   fun isStrongboxAvailable(context: Context): Boolean {
-    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/oblador/keychain/DeviceAvailability.kt
```diff
@@ -15,7 +15,7 @@ import androidx.biometric.BiometricManager
 object DeviceAvailability {
 
   fun isStrongboxAvailable(context: Context): Boolean {
-    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
```
androideveloper, line 18: This will actually make the `supportsSecureHardware` return false, but actually it's still hardware backed with TEE

PR #755, opened 2025-05-16
feat(android): add confirmationRequired option to authenticationPrompt

quangsuong:
### ✨ Summary

This PR introduces a new optional field `confirmationRequired` in the `authenticationPrompt` config for Android. It allows developers to control whether the user must press an explicit confirmation (OK) button after biometric authentication.

By default, `confirmationRequired` is now set to `false`, enabling a faster, more seamless biometric UX unless explicitly overridden.

---

### ✅ Changes

**Android Native:**
- `KeychainModule.kt`:
  - Reads `confirmationRequired` from `authenticationPrompt`.
  - Applies it to `PromptInfo.Builder.setConfirmationRequired(...)`.
  - **Default is now `false`** (user does **not** need to confirm after authentication unless specified).

**TypeScript Types:**
- Extended `AuthenticationPrompt` type to support:

```ts
confirmationRequired?: boolean;

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.kt
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.kt
@@ -68,6 +68,7 @@ class KeychainModule(reactContext: ReactApplicationContext) :
       const val SUBTITLE = "subtitle"
       const val DESCRIPTION = "description"
       const val CANCEL = "cancel"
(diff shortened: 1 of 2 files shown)


DorianMazur:
@herbig Well, that's not really true. It defaults to true
https://developer.android.com/reference/androidx/biometric/BiometricPrompt.PromptInfo.Builder#setConfirmationRequired(boolean)

Issue #36, opened 2016-10-19
How do i store other info in keychain except username and password


hejiaji:
could it support to store other custom property like 'token', 'certification' or any other info rather than username and password?

As i see now it only supports these two properties which is a restriction to us

hejiaji:
@oblador @vonovak

vonovak:
You can serialize what you need to json and save it as the password. Then
you just deserialize the password.

hejiaji:
Ok, makes no sense to me, but seems like it's a only way

dannyharding10:
I would love to have a feature like this. Storing key-value pairs would be a huge plus. I wish there was a function like `Keychain.setValueForKey('myKey', 'myValue')`, and then you can get that value with something like `Keychain.getValueForKey('myKey')`. Right now you have to store pairs of data of usernames and passwords which is not optimal for simply storing a token or a password by itself.

hejiaji:
@dannyharding10 That's the exact feature i wanted

leomeloxp:
Is there an actual practical difference between the two?

- `Keychain.setGenericPassword(username, password)`
- `Keychain.setValueForKey('myKey', 'myValue')`

I believe you can achieve the same end result by just using the generic password approach. I'm interested to know any challenges you find with this though, as this is one of the use cases I'm about to implement on an app.

dannyharding10:
One challenge would be if your app interfaced with multiple APIs and it had to store a token for each one.  You could  only store one or two in the generic password.  There would be ways to get around this, but they're not great solutions because it would make the code hard to understand.  I ended up using the generic password myself because I only had to store one token, but this is still not optimal because generic password is meant for storing a username and a password pair.

danielmartinprieto:
I agree with you, guys. It'd be great to have a simple API to store a key-value pair.

MacKentoch:
As answered @vonovak you can just serialize your data (*like JSON.stringify*) to store any key-value data.

danielmartinprieto:
Hey, @MacKentoch, thanks for answering.

My problem is not storing an object, but to store some info (a string, a stringified object...) under a certain key so I can retrieve it later, easily. What I'm doing right now is:
```javascript
Keychain.setInternetCredentials('KEY', 'RANDOM_STUFF', 'value').then(() => console.log('Done'))
Keychain.getInternetCredentials('KEY').then(({ password }) => console.log('Got it', password))
```
but ideally we need
```javascript
[...]

PR #765, opened 2025-08-18
feat: Keychain items array

mattlennon3:
This MR makes it possible to store an array of items in the keychain.

**Reason for MR**

I am raising this MR because I wanted to migrate our project from using `react-native-sensitive-info` to using `react-native-keychain`. This repo is actually maintained and has a larger community using it. Also the sensitive info library seems to stop working after upgrading RN `0.76.1+` https://github.com/mCodex/react-native-sensitive-info/issues/424#issuecomment-2511187606. So, time to move on!

However, during testing I found that this library only returns a single object from the keychain. Which seems to be intended looking at comments here: https://github.com/oblador/react-native-keychain/issues/36#issuecomment-329202738, I should be serialising my data structure and storing it as a string in this single object.

This would be fine if I was starting a new project, but losing access to data already in the keychain was not acceptable as we would lose important login keys etc. So I decided to convert the functionality over to this library. Hopefully this will help others convert to this library and add some nice new functionality for existing users.

**Changes**

I have added these functions: `getItemForKey` `getAllItems` `setItemForKey` `removeItemForKey` `clearItems`.

Of course a requirement for me was to mirror as best as possible the behaviour of the functions in `react-native-sensitive-info`. So I can preserve my data after migrating to this library.

I feel like the iOS implementation fits in nicely with the existing code. But I will admit that the Android side was more complex, you guys have a `DataStorePrefsStorage` class that I wasn't sure how exactly to integrate with, so I played it safe and kept to a simple implementation within the main `KeychainModule`. Would appreciate any pointers for this.

Thanks and any feedback is welcome 🙏

### DIFF
--- a/android/src/main/java/com/oblador/keychain/KeychainModule.kt
+++ b/android/src/main/java/com/oblador/keychain/KeychainModule.kt
@@ -446,10 +446,8 @@ class KeychainModule(reactContext: ReactApplicationContext) :
       val value = prefs.getString(key, null)
       
       if (value != null) {
-        // Return the value directly to match old format
         promise.resolve(value)
       } else {
-        // Return undefined/null when key doesn't exist
         promise.resolve(null)
       }
     } catch (e: Exception) {
@@ -491,7 +489,6 @@ class KeychainModule(reactContext: ReactApplicationContext) :
       editor.putString(key, value)
       editor.apply()
       
-      // Return the value to match old format
       promise.resolve(value)
     } catch (e: Exception) {
       Log.e(KEYCHAIN_MODULE, "Error setting item for key: ${e.message}", e)
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,7 +17,6 @@ import type {
   SetOptions,
   AuthenticationTypeOption,
   AccessControlOption,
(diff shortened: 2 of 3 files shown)


mattlennon3:
Hi @DorianMazur , thanks for the response. Sorry I've not added any README changes or other documentation yet, I've had to put my upgrade on hold to fix some other issues. But I'm ramping back up now.

To your question, because of the limited retrieval functionality I believe a migration step is not possible, as I alluded to in my description:
> However, during testing I found that this library only returns a single object from the keychain.

I am referring to `getGenericPassword` here, which only returns the first item in the array of saved passwords. If this library had a `getAllPasswords` type function, I could write a migration step, but it doesn't. 

There is also no period where we can have both `react-native-sensitive-info` and this library installed. We have to take a step from the old library, to this new one. Because of this:
> the sensitive info library seems to stop working after upgrading RN 0.76.1+

So if you want, I could refactor this PR to have the ability to return the array of passwords, but not set them. Then a migration step is possible. Or we can keep the full functionality of updating the array of passwords, which I do think is valid as it prevents this arguably "heavy" serialising/deserialising of data.

Thanks

